PR #22, opened 2015-07-27
[preview] dataset api

### LINKED ISSUE #7: Dataset support
https://github.com/mapbox/mapbox-dataset-sdk and https://github.com/mapbox/mapbox-data-cli both implement interfaces to the upcoming dataset API.

We should integrate the best of both into this library.

rclark:
- adds work-in-progress dataset api functions to the Mapbox client
- decodes the requesting user's name from the provided access token
- adds tests to fully cover branching in `make_service.js`
- `.gitignore` coverage outputs

Additional scopes are needed to pass tests on travis. fixes #7 

cc @tmcw @willwhite


lib/get_user.js
```diff
@@ -0,0 +1,20 @@
...
+
+  try {
+    data = (new Buffer(data, 'base64')).toString('utf8');
+    return JSON.parse(data).u;
+  } catch(err) {
+    return null;
+  }
+}
+
+module.exports = getUser;
```
willwhite: Any benefit to including code versus just parsing using a jwt library?

lib/constants.js
```diff
@@ -6,3 +6,7 @@ module.exports.API_GEOCODER_REVERSE = compile('${endpoint}/v4/geocode/${dataset}
 module.exports.API_DIRECTIONS = compile('${endpoint}/v4/directions/${profile}/${encodedWaypoints}.json?${query}');
 module.exports.API_SURFACE = compile('${endpoint}/v4/surface/${mapid}.json?${query}');
 module.exports.API_MATCHING = compile('${endpoint}/matching/v4/${profile}.json?${query}');
+module.exports.API_DATASET_DATASETS = compile('${endpoint}/datasets/v1/${user}?${query}');
```
willwhite: We have been trying to avoid `user` in this context because it is ambiguous with the user that's authenticating the request. Could we use `{owner}` instead?

tmcw:
Looks pretty good: general question: is it a covered usecase for one user to pull/push from another user's datasets? Right now this assumes that token=user, but will that assumption continue to be true?

lib/services/datasets.js
```diff
@@ -0,0 +1,528 @@
...
+ * //     }
+ * //   ],
+ * //   "delete": [
+ * //     "feature-id-1",
+ * //     "feature-id-2"
+ * //   ]
+ * // }
+ * });
+ */
+Datasets.prototype.bulkFeatureUpdate = function(put, deletes, dataset, callback) {
```
tmcw: could this be `puts` to match the plural of `deletes`?

lib/get_user.js
rclark: When I looked for modules I wasn't aware that browserify does some polyfilling, so I was trying to find something that was node + browser. The client side libs for this I found are **way** overkill for what ends up being incredibly straightforward code. I see no reason to bring in a dependency to do this.

lib/services/datasets.js
rclark: Yes, I'll work this in. I want to make sure that the names of the variables here match the properties of the response object returned from the API.

rclark:
About other people's datasets: I don't believe that it is currently possible, but it is something that we want to make possible. Datasets' ids make this a little bit clumsy. URLs are like:

```
GET /datasets/v1/:owner/:datasetid/features?access_token=:token
```

In the context of requesting your own dataset, you configure the SDK with your access token, and so you only need to provide the `:datasetid` in order to find your features. In a sense, though, you can see that the dataset's id _is not_ just `:datasetid`, really it is `:owner/:datasetid`. 

Where a dataset id is required to make a request, a method in this lib looks like

```js
datasets.readFeature = function(featureid, datasetid, callback) {
  // GET /datasets/v1/:user-from-token/:datasetid/feature/:featureid?access_token=:token
};
```

Even though it is honest, I kinda don't want to change this to

```js
datasets.readFeature = function(featureid, owner, datasetid, callback) {};
```

... because much of the time the owner will be you and feel redundant. What if instead, add logic to each function that takes a dataset id:

```js
var owner;
var id;
if (datasetid.split('/').length > 1) {
  owner = dataset.id.split('/')[0];
  id = dataset.id.split('/')[1];
} else {
  owner = this.user; // was determined via access token
  id = dataset;
}
```

lib/constants.js
rclark: Sounds good -- are you cool in code referring to the person making the request as the `user` then?

willwhite: :+1:

lib/services/datasets.js
```diff
@@ -0,0 +1,528 @@
...
+ * var mapboxClient = new MapboxClient('ACCESSTOKEN');
+ * mapboxClient.listDatasets(function(err, datasets) {
+ *   console.log(datasets);
+ *   // [
+ *   //   {
+ *   //     "owner": {account},
+ *   //     "id": {dataset id},
+ *   //     "name": {dataset name},
+ *   //     "description": {dataset description},
+ *   //     "created": {timestamp},
```
willwhite: Feel free to defer this to another conversation, but I'm curious if you have thought about documenting return types/properties with documentationjs. By using placeholders in this "example", we miss an opertunity to communicate specifics about types or return formats - like the fact that the `created` property is a ISO formatted string.

Ideally:

- return types would be thoroughly documented
- examples would be pure examples and show real looking values - not place holders

tmcw: Upstream in JSDoc, there's no support for documenting [parts of return values](http://usejsdoc.org/tags-returns.html) like you can [document parts of params](http://usejsdoc.org/tags-param.html). See previous discussion: https://github.com/documentationjs/documentation/issues/127

rclark: I'm 100% with you on the ideal case here. But `@returns` just isn't the right place to put this. What you really want to do is document the `@param {Function} callback`, which is a part of JSDoc that seems kinda clumsy and isn't supported in documentationjs. 

I do think this is something we ought to elevate to the shortlist in documentationjs: https://github.com/documentationjs/documentation/issues/9

tmcw: @rclark do you mean the `@callback` tag?

rclark: Yep @tmcw that's what I was getting at.

lib/services/datasets.js
```diff
@@ -0,0 +1,528 @@
...
+ *   // }
+ * });
+ */
+Datasets.prototype.updateDataset = function(dataset, options, callback) {
+  invariant(typeof dataset === 'string', 'dataset must be a string');
+  invariant(typeof callback === 'function', 'callback must be a function');
+  invariant(typeof options === 'object', 'options must be an object');
+  invariant(!!options.name || !!options.description, 'options must include a name or a description');
+
+  if (options.name) invariant(typeof options.name === 'string', 'options.name must be a string');
```
willwhite: Should we do this type of validation here or let it happen on the server? I could see us making a backward compatible change that removes this requirement.

rclark:
I'm happy with this for now. Aside from the cross-owner dataset access thing which I'd like to punt on until we're more clear on how we'll implement it, I think that I hit all the issues that you guys spotted. Good to merge @tmcw @willwhite?

tmcw:
:+1: yep!

willwhite:
step on it :+1:

PR #24, opened 2015-07-29
get_user helper to determine owner of the provided access_token

willwhite:
Load user from the provided access_token and set it on the request client. Service libs in the future will use this as a default for requests.

/cc @rclark

### DIFF
--- a/lib/get_user.js
+++ b/lib/get_user.js
@@ -2,6 +2,7 @@
 
 function getUser(token) {
   var data = token.split('.')[1];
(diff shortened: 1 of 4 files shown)


lib/get_user.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+function getUser(token) {
+  var data = token.split('.')[1];
```
tmcw: this will throw if the token does not contain a `.`

PR #26, opened 2015-07-30
Add uploads

willwhite:
Adds support for the [uploads api](https://www.mapbox.com/developers/api/uploads/).


tmcw:
How do you feel about the S3 part of this? Should mapbox-sdk-js require aws-sdk and manage this for you, or should we just provide code samples that demonstrate integration?

willwhite:
I'm learning towards keeping this a very light wrapper around the HTTP API (and small dependency tree especially for the browser). It seems like something higher level that could be added later or by another project. That said, I could be convinced either way.

rclark:
> keeping this a very light wrapper around the HTTP API

:+1: IMO this SDK should provide the base-level building blocks and let service-specific tools handle higher-level niceties where needed.

tmcw:
Yep, agree. The aws-sdk is also a very heavy dependency and not all users are going to use uploads.

PR #27, opened 2015-07-30
Allow override of account during service construction

willwhite:
Super low-impact way of handling cross account access.

/cc @rclark


rclark:
:+1: :+1:

rclark:
Actually just back to this semantic issue you've brought up before: since this is actually letting the requesting user set the owner of the resources that will be requested, shouldn't we call the property `owner` instead of `user`?

PR #28, opened 2015-08-14
Add browser test script

tmcw:
This adds smokestack and enough polyfill to get node-tap working
browser-side. The tests run, but the upload tests fail partially
through and this new test environment exposes a superagent oddity:
that it's very weak in terms of explicitly parsing responses
into JSON.

Passing to @scothis to test out `rest` as a better replacement
to superagent in this context.

### DIFF
--- a/lib/services/datasets.js
+++ b/lib/services/datasets.js
@@ -8,6 +8,9 @@ var invariant = require('invariant'),
   constants = require('../constants'),
   makeURL = require('../make_url');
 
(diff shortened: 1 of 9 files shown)


scothis:
I'm not quite sure what's going on, but when running these tests, the node process get's stuck consuming 100% cpu. Killing the node process allows the tests to continue until is happens again. Had to kill 5 processes before the test suite completed.

tmcw:
Superseded by #29

PR #29, opened 2015-08-17
Replace superagent with rest

scothis:
This is a minimal replacement of superagent with rest. There are a number of way to take deeper advantage of rest.js:
- use `rest/interceptor/pathPrefix` to manage the api base
- use path param replacement instead of es6 template strings (it's smarter about url encoding and query string construction)
- use `rest/interceptor/defaultRequest` to set the access_token universally
- be smarter about error messages emitted
- return promises from the api methods, making callbacks optional
- create a mime serializer that auto lints geojson

There are a number of test failures that still need to be resolved.


tmcw:
This PR is super close to landing, and I'm :100: convinced that `rest` is a high-quality library. Have one final issue to work through, in which the order of the templating and default interceptors seems to matter and one is blowing away params that we need to land in the request.

scothis:
The template interceptor will consume all of the request params, even if they are not defined on the template. It also requires that each param that can be used to exist in the template. Unlike the default param support which will apply path params and append leftovers on the query string.

scothis:
@tmcw @rclark `npm test` now passes `npm run browser-test` is still funky.

tmcw:
:+1: This is ready for a merge. Browser compatibility deserves its own ticket.

PR #30, opened 2015-08-17
nullcheck for geocoder

jonching:


tmcw:
#29 covers this as well

PR #32, opened 2015-08-24
Tilestats

rclark:
Adds support for reading tileset statistics.

This is branched off from #29, which will need to merge before this can go out the door.


lib/constants.js
```diff
@@ -14,3 +14,6 @@ module.exports.API_DATASET_DATASETS = '/datasets/v1/{owner}';
 module.exports.API_DATASET_DATASET = '/datasets/v1/{owner}/{dataset}';
 module.exports.API_DATASET_FEATURES = '/datasets/v1/{owner}/{dataset}/features';
 module.exports.API_DATASET_FEATURE = '/datasets/v1/{owner}/{dataset}/features/{id}';
+module.exports.API_TILESTATS_STATISTICS = '{endpoint}/tilestats/v1/{owner}/{tileset}';
```
scothis: The endpoint is automatically applied to the URL, you don't need to include it here.

rclark: :+1: got it.

PR #38, opened 2015-09-03
Add Distance API support. Fixes #36

### LINKED ISSUE #36: Support distances api

tmcw:
cc @freenerd for the review

### DIFF
--- a/lib/services/distance.js
+++ b/lib/services/distance.js
@@ -1,0 +1,83 @@
+'use strict';
+
+var invariant = require('invariant'),
+  makeService = require('../make_service'),
+  constants = require('../constants');
+
+var MapboxDistance = makeService('MapboxDistance');
+
+/**
+ * Compute a distance table of all paths between a set of waypoints.
+ * Consult the [Mapbox Distance API](https://www.mapbox.com/developers/api/distance/)
+ * for more documentation.
+ *
+ * @param {Array<Array<number>>} waypoints an array of objects with `latitude`
+ * and `longitude` properties that represent waypoints in order. Up to
+ * 100 waypoints can be specified.
+ * @param {Object} [options={}] additional options meant to tune
+ * the request
+ * @param {string} [options.profile=driving] the directions
+ * profile, which determines how to prioritize different routes.
+ * Options are `'driving'`, which assumes transportation via an
+ * automobile and will use highways, `'walking'`, which avoids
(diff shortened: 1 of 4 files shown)


lib/services/distance.js
```diff
@@ -0,0 +1,83 @@
...
+
+var MapboxDistance = makeService('MapboxDistance');
+
+/**
+ * Compute a distance table of all paths between a set of waypoints.
+ * Consult the [Mapbox Distance API](https://www.mapbox.com/developers/api/distance/)
+ * for more documentation.
+ *
+ * @param {Array<Array<number>>} waypoints an array of objects with `latitude`
+ * and `longitude` properties that represent waypoints in order. Up to
```
freenerd: it's `[ longitude, latitude ]`

lib/services/distance.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+var invariant = require('invariant'),
+  makeService = require('../make_service'),
+  constants = require('../constants');
+
+var MapboxDistance = makeService('MapboxDistance');
+
+/**
+ * Compute a distance table of all paths between a set of waypoints.
```
freenerd: note that the distance table returns travel times and not actual distances?

lib/services/distance.js
```diff
@@ -0,0 +1,83 @@
...
+ *   console.log(results);
+ * });
+ *
+ * // Results is an object like:
+ * { durations:
+ *   [ [ 0, 1196, 3977, 3415, 5196 ],
+ *     [ 1207, 0, 3775, 3213, 4993 ],
+ *     [ 3976, 3774, 0, 2650, 2579 ],
+ *     [ 3415, 3212, 2650, 0, 3869 ],
+ *     [ 5208, 5006, 2579, 3882, 0 ] ] }
```
freenerd: for the effect of demonstration include a `null` value here?

freenerd:
It the api returns a non-200 code, will that show up in the `err` param of the callback? The API returns a fair share of 40X for invalid accessToken, invalid coordinates etc with a descript body.

lib/services/distance.js
tmcw: :+1: added a null example

tmcw:
Any JSON error responses returned by the server will be available as the `err` argument to the callback

PR #39, opened 2015-09-03
Split services by GA and flagged.

### LINKED ISSUE #14: Add API access limitations to README and doc
To get access to Directions and Map Matching, users need to:

- have a paid account
- email help@mapbox.com so we can check the flag for their accounts

This should be in the README and the doc. Or just the doc?

tmcw:
cc @lyzidiamond - the right things are in the right buckets, right?

Fixes #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,20 +9,28 @@ to Mapbox services.
 
 ## Services
 
+Generally Available
+
 * [Geocoding](https://www.mapbox.com/developers/api/geocoding/)
   * Forward (place names ⇢  longitude, latitude)
   * Reverse (longitude, latitude ⇢ place names)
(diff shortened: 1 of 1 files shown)


lyzidiamond:
@tmcw yes, but... right now the way flags work on the admin side, the "directions" flag on accounts handles map matching, distance, and directions, even though directions should be fully available to everyone without a flag. (Surface has its own flag.) Is this correct @geografa? If so, we should split out the flags and open a ticket for that.

lyzidiamond:
Okay, revision to above. Per https://github.com/mapbox/api-directions/pull/284, _every user_ has access to directions, and the `directions` flag adds access to map-matching and distance. Confusing, but at least it works how we want it to.

Disregard above comment, this is good to go @tmcw :+1:

PR #46, opened 2015-09-18
Add support for Static Maps API. Fixes #45

### LINKED ISSUE #45: Adding a Static Service
I'm looking into adding a `static` [service](https://github.com/mapbox/mapbox-sdk-js/tree/master/lib/services) to this library for generating URLs to static map images.

Here's a rough API sketch: 

```js
var MapboxClient = require('mapbox');
var client = new MapboxClient('YOUR_ACCESS_TOKEN');
var url = client.static('mapbox.pencil', ({lon: 123, lat: 456});
```

Is the above feature appropriate for this library?

I can't seem to find a service method that doesn't take a callback. Are there any, or do all the methods hit the network?

tmcw:
cc @zeke - see test/services/static.js for how the API works

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -18,3 +18,4 @@ module.exports.API_DATASET_FEATURE = '/datasets/v1/{owner}/{dataset}/features/{i
 module.exports.API_TILESTATS_STATISTICS = '/tilestats/v1/{owner}/{tileset}';
 module.exports.API_TILESTATS_LAYER = '/tilestats/v1/{owner}/{tileset}/{layer}';
 module.exports.API_TILESTATS_ATTRIBUTE = '/tilestats/v1/{owner}/{tileset}/{layer}/{attribute}';
+module.exports.API_STATIC = '/v4/{mapid}{+overlay}/{+xyz}/{width}x{height}{+retina}{.format}';
--- a/lib/services/static.js
+++ b/lib/services/static.js
@@ -44,7 +44,7 @@ MapboxStatic.prototype.getStaticURL = function(mapid, width, height, position, o
 
   var defaults = {
     format: 'png',
-    retina: false
+    retina: ''
   };
 
   var xyz;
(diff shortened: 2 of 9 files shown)


lib/constants.js
```diff
@@ -18,3 +18,4 @@ module.exports.API_DATASET_FEATURE = '/datasets/v1/{owner}/{dataset}/features/{i
 module.exports.API_TILESTATS_STATISTICS = '/tilestats/v1/{owner}/{tileset}';
 module.exports.API_TILESTATS_LAYER = '/tilestats/v1/{owner}/{tileset}/{layer}';
 module.exports.API_TILESTATS_ATTRIBUTE = '/tilestats/v1/{owner}/{tileset}/{layer}/{attribute}';
+module.exports.API_STATIC = '/v4/{mapid}{+overlay}/{+xyz}/{width}x{height}{+retina}{.format}';
```
scothis: you can treat `x`, `y` and `z` as discrete values with `{x,y,z}`

tmcw: this url segment can be x,y,z or it can be `auto`, so i made it one chunk instead of three

lib/services/static.js
```diff
@@ -0,0 +1,97 @@
...
+    mapid: mapid,
+    width: width,
+    xyz: xyz,
+    height: height
+  });
+
+  if (params.retina) {
+    params.retina = '@2x';
+  } else {
+    params.retina = '';
```
scothis: you can leave `params.retina` undefined, unless you need to overwrite

lib/constants.js
scothis: For 'auto' what you have is what you want

zeke:
Hey sorry I didn't chime in earlier. This looks great. Thanks for your responsiveness!

zeke:
I finally got around to using this. Works like a charm. Thanks again, @tmcw.

I'll keep my eye out for the next release so we can use npm instead of github master.

tmcw:
Cool: next release should be soon, right now we're making some tweaks to directions api to get it working with sdk-js, then getting ci tests to run in browsers, and then there'll be a release.

PR #52, opened 2015-10-05
Bump geocoding to v5

heyitsgarrett:


bsudekum:
One would think this would break tests.

heyitsgarrett:
This actually shouldn't break legacy tests -- v5 should be interchangeable with v4 as far as results are concerned. Right now, tests are failing on the `uploads.js` test, which seems to be unrelated. Digging in on that now.

PR #55, opened 2015-10-14
Make it possible to specify GET params, and fix geocoder proximity

### LINKED ISSUE #54: Geocoder proximity doesn't do anything
The forward geocoder endpoint passes a proximity flag into the rest client, but it doesn't make it into the final URL, so it doesn't end up affecting the results. The issue seems to be [this rest interceptor](https://github.com/mapbox/mapbox-sdk-js/blame/master/lib/client.js#L15), which, per the [rest docs](https://github.com/cujojs/rest/blob/master/docs/interceptors.md#module-rest/interceptor/template), does this:

> Params defined by the template that are missing as well as additional params are ignored. After the template interceptor, the request.params are removed from the request object, as the URI is fully defined.

So the object is gone by the time GET parameters are added.

Assigning to myself, but not sure yet how to fix it.

cc @tmcw

apendleton:
The rest template interceptor deletes params after template interpolation, so params intended to be GET parameters never get used. This adds a separate "getParams" option that the template interceptor doesn't touch, and makes the geocoder use it, to fix #54 .

If we don't want to go this direction, the other options are either to not use the template interceptor at all and depend on UriBuilder instead, which doesn't have this problem but is apparently deficient in other ways as suggested by the rest docs, or to fork the template interceptor and change its behavior in this respect.

cc @tmcw

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2,6 +2,7 @@
 
 var rest = require('rest');
 var callbackify = require('./callbackify');
+var interceptor = require('rest/interceptor');
 
 // rest.js client with MIME support
 module.exports = function(config) {
@@ -12,6 +13,11 @@ module.exports = function(config) {
     .wrap(require('rest/interceptor/defaultRequest'), {
       params: { access_token: config.accessToken }
     })
+    .wrap(interceptor({ request: function(request, config) {
+        // if there are any explicitly-marked get parameters, stick them in 'params' now that interpolation is done
+        if (request.getParams) request.params = request.getParams;
+        return request;
+     }}))
     .wrap(require('rest/interceptor/template'))
     .wrap(callbackify);
 };
--- a/lib/services/geocoder.js
+++ b/lib/services/geocoder.js
@@ -53,12 +53,13 @@ MapboxGeocoder.prototype.geocodeForward = function(query, options, callback) {
     query: query,
     dataset: 'mapbox.places'
   };
+  var getOptions = {};
 
   if (options.proximity) {
     assert(typeof options.proximity.latitude === 'number' &&
       typeof options.proximity.longitude === 'number',
       'proximity must be an object with numeric latitude & longitude properties');
-    queryOptions.proximity = options.proximity.longitude + ',' + options.proximity.latitude;
+    getOptions.proximity = options.proximity.longitude + ',' + options.proximity.latitude;
   }
 
   if (options.dataset) {
@@ -69,6 +70,7 @@ MapboxGeocoder.prototype.geocodeForward = function(query, options, callback) {
   this.client({
     path: constants.API_GEOCODER_FORWARD,
     params: queryOptions,
+    getParams: getOptions,
     callback: callback
   });
 };


tmcw:
I'm confused by this - are all query parameters being removed by the template interceptor? cc @scothis

apendleton:
@tmcw yup: https://github.com/cujojs/rest/blob/master/interceptor/template.js#L43

I think because the template interceptor somewhat overlaps in functionality with the final-step UriBuilder invocation, but UriBuilder assumes that if there are things that are in params that don't have slots in the template, they're supposed to be GET parameters. Since the template interceptor has already filled in all the slots, if you comment out that delete, they all reach UriBuilder with no slots to go in and all get added as GET parameters, even if some of them have already been interpolated into the URL (so concretely in the reverse geocoding example, you get latitude and longitude both in the path and in the GET params).

The proper behavior should probably be that the template interceptor only deletes the params that are consumed, rather than the whole object, but making it work like that would require refactoring both the interceptor and https://github.com/cujojs/rest/blob/master/util/uriTemplate.js

I'd be fine with that, but I'm not sold on the correctness of conflating path components with GET parameters anyway, so I don't think making them separately settable is the end of the world. Plus it's easier.

apendleton:
I should qualify: all params that were set before the template interceptor are removed by the template interceptor. There's another interceptor that adds in the access token, and that one works because it runs after the template interceptor.

scothis:
@apendleton exactly.

Automatic template expansion in the root client was the one bit of magic I put into rest.js and it's biting us now. Won't make that mistake again...

The best thing to do is add the params to the URI template if they are known, if a param is in the template without a value, it's ignored. For example, to populate the query string for props `a`, `b`, and `c`:
- `/some/path{?a,b,c}`
- `/some/path?existing=query{&a,b,c}`

If you don't know what params would be included, a custom interceptor deeper then the template interceptor is needed to repopulate the request params from another source.

apendleton:
@scothis I'm currently testing this by situating a proxy between the client and the server and watching URLs go by, and adding {?proximity} to the end of the template URL doesn't make the proximity show up in the query string in the proxy logs. I can push that to another branch if you want to confirm.

This PR is basically your option number 2 (adding another interceptor that adds the values back in).

apendleton:
@scothis whoops, scratch that last comment. Typo on my part. That works.

apendleton:
Okay, I'm gonna close this, nuke this branch, and do that instead. In general, though, this kind of silent failure is kind of concerning. Current master passes all of its tests, including tests of this option, even though it doesn't work, and there's no easy way to tell it's not working other than the proxy hack. I only happened to notice because I'm trying to change the behavior of the proximity and it was the only way I could figure out to check that what I was doing was working. I don't see an obvious way to get the constructed URL out, so I'm not even sure how to write a test against this happening again in the future.

tmcw:
@apendleton the tests are intentionally lax on specific result JSON because that can change quickly on the API level and may even be non-deterministic. For the proximity option, we could confirm it's working by looking at the results and making sure they're all in the radius.

scothis:
For offline testing the best thing to do is swap in a mock [default client](https://github.com/cujojs/rest/blob/master/docs/clients.md#default-client).

```
rest.setDefaultClient(function (request) {
  var response = { request: request };
  // set other properties on the response as needed
  return response;
});
```

The next release of rest.js will include the requested url on the response.

apendleton:
@tmcw proximity doesn't set a radius into which results will fall, it just supplies a hint that might affect the results by promoting what otherwise would not have been the top hit to be the top hit if you happen to be close to it. So, "San Francisco" generally returns San Francisco, CA, unless you happen to be in Argentina, and then you get [this San Francisco](https://en.wikipedia.org/wiki/San_Francisco,_C%C3%B3rdoba) instead. Most of the time it won't affect results at all, and whether it does or doesn't will, as you say, change as data chances, so I don't think there's a reliable way to test the result JSON; in any event, I think testing whether or not the proximity parameter does what it's supposed to is the provence of the test suite of the geocoder, not the API wrapper. All I'd want to do in the wrapper is make sure it's submitting what it's supposed to, by examining the URL.

@scothis cool, I'll make a note for myself to swing back around and add a URL test when the next release is out.

PR #57, opened 2015-10-14
Configurably round reverse geocoding requests to 5 places, and proximity requests to 3

apendleton:
This nudges reverse geocodes to 5 places (~.5-meter max error) and proximity parameters in forward geocodes to 3 places (~50 meter max error) to improve performance by increasing cache hit rates, and adds corresponding tests.

cc @sbma44

### DIFF
--- a/lib/services/geocoder.js
+++ b/lib/services/geocoder.js
@@ -6,6 +6,9 @@ var assert = require('assert'),
 
 var MapboxGeocoder = makeService('MapboxGeocoder');
 
+var REVERSE_GEOCODER_PRECISION = 5;
+var FORWARD_GEOCODER_PROXIMITY_PRECISION = 3;
+
 /**
  * Search for a location with a string, using the
  * [Mapbox Geocoding API](https://www.mapbox.com/developers/api/geocoding/).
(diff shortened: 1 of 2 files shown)


sbma44:
Looks good to me. Mind throwing a linter at it? I think I might've made some formatting errors in the code I sent you.

apendleton:
@sbma44 nope, you were good. The only eslint failure was mine.

lib/services/geocoder.js
```diff
@@ -119,15 +128,26 @@ MapboxGeocoder.prototype.geocodeReverse = function(location, options, callback)
...
-      latitude: location.latitude,
+      longitude: roundTo(location.longitude, precision),
+      latitude: roundTo(location.latitude, precision),
       dataset: dataset
     },
     callback: callback
   });
 };
 
+function roundTo(value, places) {
```
tmcw: Is this roughly equivalent to using [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed)?

sbma44: it's pretty close, but I don't think it's desirable to pad with extra zeroes (e.g. `2.00000`) for cache hit rate purposes.

tmcw:
:+1: lgt,

test/geocode.js
```diff
@@ -59,6 +59,26 @@ test('MapboxClient#geocodeForward', function(t) {
...
+      opts.params.proximity.split(',').forEach(function(coord, i) {
+        t.ok((coord.split('.')[1].length <= 3), 'proximity coordinate [' + i + '] precision <= 3');
+      });
+      t.ok(opts.params.proximity, 'proximity is set')
+      opts.callback();
+    }};
+
+    t.ok(client);
+    client.geocodeForward.apply(tester, ['Paris', {
+      proximity: { latitude: 33.6875431, longitude: -95.4431142 }
```
tmcw: it looks like this should specify a `precision` option but it doesn't? the opts.precision handling is not covered by tests in this branch.

lib/services/geocoder.js
```diff
@@ -119,15 +128,26 @@ MapboxGeocoder.prototype.geocodeReverse = function(location, options, callback)
     dataset = options.dataset;
   }
 
+  var precision = REVERSE_GEOCODER_PRECISION;
+  if (options.precision) {
+    assert(typeof options.dataset === 'number', 'precision option must be number');
```
tmcw: this should be options.precision

tmcw:
I'll fix the inline comments.

Issue #59, opened 2015-11-05
better uri encoding for geocodes


samanpwbb:
I was doing a bit of [forward geocoding](https://github.com/mapbox/mapbox-sdk-js/blob/master/API.md#geocodeforward) and ended up running a string with a line break in it, looked sort of like `100 6th St\nSan Francisco`. It failed with this error message:
 
```
Potentially unhandled rejection [7] Error: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<HTML><HEAD><META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=iso-8859-1">
<TITLE>ERROR: The request could not be satisfied</TITLE>
</HEAD><BODY>
<H1>ERROR</H1>
<H2>The request could not be satisfied.</H2>
<HR noshade size="1px">
Bad request.
<BR clear="all">
<HR noshade size="1px">
<PRE>
Generated by cloudfront (CloudFront)
Request ID: 0C0WjIdiUNxnFnej1T_D3zhMfAQMXe49QKjAXEdziNroPxHra7E3Zg==
</PRE>
<ADDRESS>
</ADDRESS>
</BODY></HTML>
```

https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/services/geocoder.js#L56

samanpwbb:
I tried encoding with like `encodeURIComponent(query)` but that failed as well. `query.replace(/(\r\n|\n|\r)/gm,'')` works, but seems like this is a problem that needs to be solved further upstream?

tmcw:
Yeah, potentially upstream in the geocoder cc @ingalls

ingalls:
@tmcw @samanpwbb Definitely seeing this error, from my tests this appears to be a problem with how the query is being encoded in mapbox-sdk-js and not api-geocoder. Since the error is being returned by cloud-front its not even hitting our server.

tmcw:
@samanpwbb can you post the code you're writing to integrate against mapbox-sdk-js? I cannot replicate the issue with the node interface of mapbox-sdk-js.

tmcw:
Upstream bug in cujo https://github.com/cujojs/rest/issues/138 cc @scothis

tmcw:
Fix landed, updating cujo

PR #60, opened 2015-11-05
Sketch bin

tmcw:
This starts to experiment with a binary interface for mapbox-sdk-js cli usage

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "0.10.0",
   "description": "interface to mapbox services",
   "main": "lib/mapbox.js",
+  "bin": {
+    "mapbox-sdk": "./bin/mapbox-sdk"
+  },
   "scripts": {
     "lint": "eslint --no-eslintrc -c .eslintrc lib test/*.js",
     "test": "tap --coverage test/browser.js test/*.js",
@@ -56,6 +59,7 @@
     "hat": "0.0.3",
     "polyline": "^0.1.0",
     "rest": "^1.3.1",
-    "xtend": "^4.0.0"
+    "xtend": "^4.0.0",
+    "yargs": "^3.29.0"
   }
 }
--- a/bin/mapbox-sdk
+++ b/bin/mapbox-sdk
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+var MapboxClient = require('../');
+var argv = require('yargs')
+  .help('help')
+  .command('geocode-forward', 'Resolve a search string to locations')
+  .demand(1)
+  .argv;
+
+if (!process.env.MapboxAccessToken) {
+  throw new Error('MapboxAccessToken environment variable must be set');
+}
+
+var mapboxClient = new MapboxClient(process.env.MapboxAccessToken);
+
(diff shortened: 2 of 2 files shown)


sbma44:
over at the python SDK we wound up going with `mbx`: https://github.com/mapbox/mapbox-sdk-py/issues/10

not sure whether or not there's a payoff to standardizing the CLI name across languages. could make scripts work more often, could make debugging more confusing for unsophisticated users.

tmcw:
It seems like having multiple binaries called `mbx` would make things really confusing given that they'd shadow each other in PATH

yhahn:
Maybe we need a cross-language package manager to arbitrate called `mpm`

sbma44:
@tmcw yeah, agreed. I should also probably note that since I wrote that comment, efforts have been made to decouple the python CLI tool and the python SDK. I'm decreasingly sure that there's a good case for bringing these kinds of patterns across languages. (emulating invocation styles/param names seems legit, to the extent it's practicable).

also obvs I strongly endorse @yhahn's suggestion

tmcw:
:+1: Also per discussion with @perrygeo I'm actually leaning toward mapbox-sdk-py being the 'official' CLI wrapper, since Python is installed by default on more systems and would be easier to get running with. I _personally_ really like having a CLI for sdk-js basically as a way to test stuff (this PR for instance was just so I could figure out #59), so I'm kind of into releasing this quietly and having it not be the user-recommended thing, but have the Python version be that.

yhahn:
`mbxjs`

and or frosted

perrygeo:
As mentioned by @sbma44 - we're in the process of moving the CLI over to a separate repo (https://github.com/mapbox/mbx-cli) in order to avoid tying the Python SDK too tightly to an official CLI. This allows

* the READMEs, docs, and branding to be adjusted for the appropriate audience
* the implementation or language behind the CLI can change in the future without disrupting user's scripts

On a related note, we're in the early stages of the CLI design so we're hoping to get some input before any initial release. @sgillies will be starting work in earnest next week so we'll track any design questions in the `mbx-cli` tickets.

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.10.0",
   "description": "interface to mapbox services",
   "main": "lib/mapbox.js",
+  "bin": {
+    "mapbox-sdk": "./bin/mapbox-sdk"
```
zeke: Why not just `mapbox`? Is that name used by some other tool?

yhahn: :trollface:

zeke: Not trolling!

yhahn: I know :smile: We actually have an internal command we've been working on called `mapbox` and we're trolling ourselves by not just swapping names (public: mapbox, internal: mbx or something else).

zeke: > public: mapbox, internal: mbx

I seeee.

karenzshea:
`mapbox-cli` namespace is all urs @tmcw

tmcw:
mapbox-cleee

tmcw:
Closing: let's dedicate efforts to https://github.com/mapbox/mapbox-cli-py !

PR #61, opened 2015-12-09
Allow pagination options if passed in listFeatures

### LINKED ISSUE #50: Feature pagination on list endpoints
We have provisional pagination in place for feature listing API endpoints. We should support these are additional options when running [`.listFeatures()`](https://github.com/mapbox/mapbox-sdk-js/blob/ffd4da71e66e010f650a8f1e1a1ea36052b925dd/lib/services/datasets.js#L207-L250). I would suggest implementing this by changing the function so that it can accept an optional `options` object that may provide the following arguments to the API call:

| Argument | Description |
| --- | --- |
| sortby | The property to use for sorting. `created` and `modified` are currently supported. |
| reverse | Set to `true` to reverse the default sort order of the listing. |
| limit | The *maximum* number of objects to return. This value must be between 1 and 100. The API will attempt to return the requested number of objects, but receiving fewer objects does not necessarily signal the end of the collection. Receiving an empty page of results is the only way to determine when you are at the end of a collection. |
| start | The object id that acts as the cursor for pagination and defines your location in the collection. This argument is exclusive so the object associated with the `id` provided to the `start` argument will not be included in the response. |

```js
Datasets.prototype.listFeatures = function(dataset, options, callback) {
  if (typeof options === 'function') {
    callback = options;
    options = {};
  }

  // ... etc
};
```

cc @isiyu @peterqliu @willwhite

tristen:
- Fixes #50


tristen:
Gah! tests sometimes pass sometimes don't. Any insight @rclark would be greatly appreciated :|

test/datasets.js
```diff
@@ -399,6 +376,109 @@ test('DatasetClient', function(datasetClient) {
...
 
+  datasetClient.test('#listFeatures', function(listFeatures) {
+    var featureIds = [];
+
+    // Setup data
+    listFeatures.test('insert some', function(assert) {
+      var client = new MapboxClient(process.env.MapboxAccessToken);
+      assert.ok(client, 'created dataset client');
+      var features = randomFeatures(3).map(function(f) {
+        f.id = hat();
```
rclark: @tristen not sure if this is the issue, but seems like if your ids are randomly generated, then you'll have a hard time making assertions about order in things like `start` and `reverse`. Maybe give the features ids 1, 2, 3 and see if the tests pass more consistently?

tristen: @rclark the line below though I am pushing it into a `featuresIds` container to keep track. But agree, this could just be more simple.

tristen:
Huh. @rclark thanks for the tip! Good here to :ship: ?

willwhite:
Nice! Thanks @tristen.

PR #63, opened 2016-01-11
add datasets back to readme with disclaimer

lyzidiamond:
About once per week we get a support request from a user using this SDK wondering why their Datasets API call 404'd. This PR adds Datasets back to the README and explains why requests to it won't work (instead of removing real methods in the SDK from `API.md`).

cc @tmcw @tristen


tristen:
:+1:

PR #77, opened 2016-03-15
Promise support sketch

### LINKED ISSUE #76: Promise support
We should support promises. This means making callbacks optional for all methods.

I'll post a sketch of this in a sec to get thoughts on the approach.

tmcw:
cc @scothis does this seem like the right approach for promise support? should we expose the full request like this and expect consumers to grab the `entity` property?

* [x] datasets.js
* [x] directions.js
* [x] distance.js
* [x] geocoding.js
* [x] matching.js
* [x] static.js
* [x] surface.js
* [x] tilestats.js
* [x] uploads.js

Fixes #76

### DIFF
--- a/lib/services/datasets.js
+++ b/lib/services/datasets.js
@@ -40,15 +40,13 @@ var Datasets = module.exports = makeService('MapboxDatasets');
  * });
  */
 Datasets.prototype.listDatasets = function(callback) {
-  invariant(typeof callback === 'function', 'callback must be a function');
-
-  this.client({
+  return this.client({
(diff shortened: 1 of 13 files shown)


scothis:
Calls to client will return a [ResponsePromise](https://github.com/cujojs/rest/blob/master/docs/interfaces.md#response-promise-methods) which is a vanilla Promise decorated with a few helper methods.

If you want the caller to have access to the full response, you can simply return the promise like you are. If you want to restrict the value to the entity you can do something like:

```
return this.client({
  ...
}).entity();
```

PR #82, opened 2016-03-24
Styles API Support. Fixes #67

### LINKED ISSUE #67: Styles API
See also https://github.com/mapbox/mapbox-sdk-py/issues/72 - we don't have, but need, API support for the Styles.

tmcw:
* [x] Retrieve a style
* [x] List styles
* [x] Create a style
* [x] Update a style
* [x] Delete a style
* [x] Retrieve font glyph ranges
* [x] Retrieve a sprite image or JSON
* [x] Add new icon to sprite
* [x] Delete icon from sprite
* [x] Embed a style

### DIFF
--- a/lib/services/styles.js
+++ b/lib/services/styles.js
@@ -1,0 +1,362 @@
+'use strict';
+
+var invariant = require('../../vendor/invariant'),
+  hat = require('../../vendor/hat'),
+  uriTemplate = require('rest/util/uriTemplate'),
+  makeService = require('../make_service'),
+  constants = require('../constants');
+
+var Styles = module.exports = makeService('MapboxStyles');
+
+/**
+ * To retrieve a listing of styles for a particular account.
+ *
+ * @param {Function} callback called with (err, datasets)
+ * @returns {undefined} nothing, calls callback
+ * @example
+ * var MapboxClient = require('mapbox');
+ * var client = new MapboxClient('ACCESSTOKEN');
+ * client.listStyles(function(err, styles) {
+ *   console.log(datasets);
+ *   // [{ version: 8,
+ *   //  name: 'Light',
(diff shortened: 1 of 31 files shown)


lib/services/styles.js
```diff
@@ -0,0 +1,220 @@
...
+ * @returns {undefined} nothing, calls callback
+ * @example
+ * var MapboxClient = require('mapbox');
+ * var fs = require('fs');
+ * var client = new MapboxClient('ACCESSTOKEN');
+ * client.addIcon('style-id', fs.readFileSync('icon.png'), function(err) {
+ *   if (!err) console.log('added icon!');
+ * });
+ */
+Styles.prototype.addIcon = function(style, iconName, icon, callback) {
```
tmcw: @scothis anything stand out as wrong in this implementation? i'm trying to send svg to the API in a PUT body, but it's being sent wrong somehow, and coming back with an invalid filetype error.

scothis: Because there isn't a custom `Content-Type` header, the mime interceptor is trying to employ the default, in this case `application/json` as [configured here](https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/client.js#L11). Add this to the request:

```
headers: {
  'Content-Type': 'application/octet-stream'
}
```

tmcw: :+1: application/octet-stream was unrecognized but test/plain works too

scothis: I should add `application/octet-stream` support by default.

If you want to use this content type it needs to be registered:
```
var registry = require('rest/mime/registry');
registry.register('application/octet-stream', registry.delegate('text/plain'));
```

... or register `image/svg+xml` to be even more 'correct'.

PR #96, opened 2016-06-12
Update API.md

jingsam:
API.md is outdated. I tried to rebuild API.md using `npm run docs`, but `xtend` was also introduced into API.md. So I manually revised the API.md


tmcw:
Thanks, we'll fix the xtend issue instead - manual changes to API.md will be overwritten by the automatic process.

PR #98, opened 2016-06-17
Upgrade to rest.js 2.0

scothis:
rest.js no longer includes a Promise library, and uses the ES6 Promise
API. A Promise polyfill from when.js is included in the vendor directory.


tmcw:
👍 awesome

PR #100, opened 2016-07-14
testing node 4 on travis

mapsam:
Adding node 4 to travis tests.

Looks like there are some unexpected issues with the token used in the tests for both `0.12` and `4`, which means this is technically failing on `master`. 

First error seems to be the culprit of the rest of subsequent errors:

```
This API requires a token with styles:list scope.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ node_js:
 - '4'
 env:
   global:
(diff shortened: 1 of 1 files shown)


mapsam:
cc @tmcw since it looks like you've worked on this most recently - thoughts around [the travis errors](https://travis-ci.org/mapbox/mapbox-sdk-js/jobs/144856677#L379)?

tmcw:
Server behavior changed with this PR by @scothis - https://github.com/mapbox/api-styles/pull/369 - tests will need a new token with the styles:list scope.

scothis:
I should be possible to mutate the scopes for an existing token inside Studio without needing to generate a new token.

mapsam:
Merging this - no need for a version bump since it just changes the travis file

PR #102, opened 2016-07-20
Modify tilestats service to fit new API

davidtheclark:
WIP in parallel with PRs to other relevant repos.

The new tilestats API is in the works, not yet deployed, so tests will fail.

The new API will have a `PUT` method, which takes a complete tilestat document, and a `GET` method, which retrieves the complete tilestat document. Dramatic simplification.

It will add a few new properties to the statistics, as well, so I updated the documentation comments.

For now I just added `#putTilestats` and modified `#getTilestats`, leaving other (legacy) methods in place. The legacy methods can then be deleted when the API no longer serves them.

cc @rclark


davidtheclark:
Clarification on this diff, which is kind of confusing: I just added a `putTilestats` method and modified the `getTilestats` method. The diff is just matching up lines in weird ways.

davidtheclark:
Now that changes to the API have been deployed to production, this PR is ready to be merged.

davidtheclark:
@tmcw and @scothis I believe that this is ready to merge and release. The new tilestats API is now in prod, and this will take advantage of it. It will PUT new stats to s3, and GET stats from s3 or fallback to dynamo. Everything else using the tilestats API should be GETting the same way right now, from s3 falling back to dynamo. Unpacker is still writing stats to dynamo; but we can't upgrade it until this merges. 

So I'm asking if you feel ok with merging this PR and cutting a release. I guess it will have to be a breaking change.

PR #104, opened 2016-08-03
Change the param `auth_token` to `api_token`

cusspvz:
Template has the variable `api_token` instead of `auth_token`
`/v4/{mapid}{+overlay}/{+xyz}/{width}x{height}{+retina}{.format}{?api_token}`

Changed to `api_token` instead so it can render the correct url.

### DIFF
--- a/lib/services/static.js
+++ b/lib/services/static.js
@@ -82,7 +82,7 @@ MapboxStatic.prototype.getStaticURL = function(mapid, width, height, position, o
     width: width,
     xyz: xyz,
     height: height,
-    access_token: this.accessToken
+    api_token: this.accessToken
   });
 
   if (params.retina === true) {


cusspvz:
Also the way the token should be injected is with: `access_token` query variable instead as described on the [documentation](https://www.mapbox.com/api-documentation/?language=cURL#access-tokens).

tmcw:
Thanks! This needs to be fixed on the other side - the constants file should reference access_token and we shouldn't waver with api_token anywhere. Pushing a fix to that end in a few seconds.

tmcw:
Fixes 2e597a4a6525fa5c61c835cf1d86049bd0aff7aa

cusspvz:
Is there a chance to release a new beta?

cusspvz:
I did my own url generator but I want to replace it with the official mapbox client. After commenting here I've tested the master branch and noticed that the current Static API client is in fact the **Classic** referenced at the docs. I've checked both *Static* and *Styles* services files and didn't found references to it, but just to be sure, does this client support **Style Static API**?

PR #106, opened 2016-08-18
Fix batchFeatureUpdate name in API docs

davidtheclark:
The API docs didn't match the actual function name: https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/services/datasets.js#L517

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,4 +1,4 @@
-# bulkFeatureUpdate
+# batchFeatureUpdate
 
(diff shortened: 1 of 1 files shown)


davidtheclark:
Oh, API.md is automatically built. And the comment in the source code is ok. So I guess this will just be fixed the next time the docs are built. Is that right @tmcw ?

davidtheclark:
I see that this has happened.

PR #111, opened 2016-10-10
Add Geocoding `Limit` Parameter

KaiBot3000:
Adds the limit parameter for geocoding calls per https://github.com/mapbox/api-geocoder/issues/1102

Tests fail locally, but the same tests are failing on master. I based this on the PR for adding bbox; please let me know if anything is missing.

### DIFF
--- a/lib/services/geocoding.js
+++ b/lib/services/geocoding.js
@@ -34,6 +34,8 @@ function roundTo(value, places) {
  * @param {string} options.types a comma seperated list of types that filter
  * results to match those specified. See https://www.mapbox.com/developers/api/geocoding/#filter-type
  * for available types.
+ * @param {number=5} options.limit is the maximum number of results to return, in range 1 to 10.
+ * Some very specific queries may return fewer results than the limit.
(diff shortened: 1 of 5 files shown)


lib/services/geocoding.js
```diff
@@ -34,6 +34,8 @@ function roundTo(value, places) {
  * @param {string} options.types a comma seperated list of types that filter
  * results to match those specified. See https://www.mapbox.com/developers/api/geocoding/#filter-type
  * for available types.
+ * @param {number=5} options.limit is the maximum number of results to return, in range 1 to 10.
```
tmcw: in _the_ range of 1 to 10? or between 1 and 10? Not sure how the grammar should function here.

KaiBot3000: Hmm, I copied the formatting of other params already there: `decimal degrees latitude, in range -90 to 90`, but agree it could be more clear. In another place I used `between 1 and 10 inclusive`.

KaiBot3000:
@tmcw This is ready for a final review. Failing tests are dataset-related.

PR #112, opened 2016-10-18
Drop the dataset batch endpoint

mcwhittemore:
As we look to move the dataset api past beta we have decided that that batch api is not something we plan to support for the long term.

This PR drops support of it fully rather than converting the batch functionality to use the single feature put and delete endpoints.

@scothis for the review


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.0.0.beta4
```
scothis: `-beta4`

lib/services/datasets.js
```diff
@@ -38,14 +38,21 @@ var Datasets = module.exports = makeService('MapboxDatasets');
...
     path: constants.API_DATASET_DATASETS,
     params: {
+      limit: opts.limit,
+      start: opts.start,
+      fresh: opts.fresh,
       owner: this.owner
     },
     callback: callback
-  }).entity();
+  });
```
scothis: Curious if @tmcw has any concern about dropping `.entity()`.

mcwhittemore:
@scothis do you think the promise response should match the callback response object?

scothis:
Yes, the callback and promise should hold the same value.

mcwhittemore:
@scothis - ok, I think this is ready now.

lib/standard_response.js
```diff
@@ -0,0 +1,18 @@
+var interceptor = require('rest/interceptor');
+
+var standardResponse = interceptor({
+  response: transform,
+});
+
+function transform(response) {
+  return {
+    url: response.url,
+    status: response.status.code,
```
scothis: It's possible some errors may not have a `status` object.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.0.0
```
scothis: Are we ready for `1.0`?

mcwhittemore: IDK? We've been on `beta` a few times now. In an earlier commit, I called this `beta4`. I can go back to that if you like.

What is our goal with 1.0?

PR #117, opened 2016-11-18
update README as Datasets API no longer in private beta

andrewharvey:
https://www.mapbox.com/blog/dataset-api-public-launch/


tmcw:
👍 Thanks Andrew!

Issue #113, opened 2016-11-07
Don't overwrite native Promise


aparshin:
I use node v4.5.0, which supports native promises.

But when I use `mapbox` module, node's native promises are replaced with polyfill from `mapbox/vendor/promise.js`

Mapbox is just a part of my project, and I want to use native promises when it is possible.

Is it possible not to overwrite native promises by default?

Test case:
```
var mapbox = require('mapbox')
console.log(Promise) //Now it is polyfill
```

tmcw:
Yep, I agree that this is much preferable. cc @scothis since this Promise implementation is there to support the `rest` library.

tmcw:
Fixed in 5d4a382c85657194cbefaa49e5f7b2f73928c291

chasemgray:
I still see this issue when loading the latest version. Angular 2 zones get a fatal error when trying to load the library.

tmcw:
Can you

scothis:
@chasemgray can you confirm you are running `1.0.0-beta5`?

Testing in the node REPL, I don't see an issue:

```
$ node
> var p = Promise
undefined
> require('mapbox')
{ [Function: service] getUser: [Function: getUser] }
> p === Promise
true
> 
```

chasemgray:
We're using JSPM for module loading. I don't think you can conditionally load when using jSPM.

davidtheclark:
The newly published `@mapbox/mapbox-sdk` prompts you to install your own Promise polyfill if you need one.

PR #122, opened 2017-01-02
Install Promise polyfill only when needed

scothis:
Switch to a new ES6 Promise polyfill that will only install itself when
the environment does not already have a native Promise implementation
available.

Issue: #113


scothis:
@chasemgray can you confirm if this branch fixes the Promise clobbering issue?

tristen:
@scothis this LGTM. Merge?

PR #129, opened 2017-02-15
update token

mapsam:
Updating travis token

cc @tristen @tmcw @scothis @mcwhittemore


mapsam:
It appears the uploads are taking a bit more time to process, so I increased the timeout on the `#readUpload` tests, which polls the uploads api for `complete: true` on the upload created previously.

PR #136, opened 2017-03-14
explain what the value of start for listFeatures

mcwhittemore:
@sarahkleins does this make things a bit clearer?


scothis:
We should remove the start property from the SDK. We now have an [easier way to load pages for listing resources](https://github.com/mapbox/mapbox-sdk-js#pagination).

As is, we're documenting the start property in this PR to be something that the API doesn't document to be true. The only [supported method for pagination](https://www.mapbox.com/api-documentation/#pagination) is via the `Link` header.

PR #148, opened 2017-04-12
fix js doc for style service

mollymerp:
copying over #141 for CI to pass ✅ 
cc @andrewharvey


mollymerp:
hmm I tested this and ran the `npm run docs` command and the `Styles` service still doesn't get added to API.md. not sure why this is happening but I will try to investigate.

PR #151, opened 2017-04-13
add editorconfig

andrewharvey:
This project uses an indent size of 2 spaces, so creating a .editorconfig file to match this

*moved from #145 for CI*


scothis:
Thanks @andrewharvey

Issue #154, opened 2017-04-13
MapboxStatic service geojson path double encoded


andrewharvey:
In the MapboxStatic service when using a `path.geojson` option, `encode_overlay` will take the GeoJSON object:
```json
{ type: 'Point', coordinates: [0, 0] }
```
JSON.stringify it:
`{"type":"Point","coordinates":[0,0]}`

encodeURIComponent it:
`%7B%22type%22%3A%22Point%22%2C%22coordinates%22%3A%5B0%2C0%5D%7D`

Which Is the correct string to return as part of the full static api call, however `uriTemplate` will then do it's own encoding on the `%`char turning it into:
`257B%2522type%2522%253A%2522Point%2522%252C%2522coordinates%2522%253A%255B0%252C0%255D%257D`

...which then fails as at the Mapbox REST API as it is invalid.

The test fixtures are incorrect as they check for the invalid double encoded string.

@scothis I tried to find a way for `uriTemplate` from `rest` to skip the encoding but it doesn't seem possible?

We could leave all the encoding to `uriTemplate` from `rest` by skipping the `encodeURIComponent` step which results in:
`%7B%22type%22:%22Point%22,%22coordinates%22:[0,0]%7D`
This works as a URL in Chrome, but breaks other clients like curl.

scothis:
@andrewharvey yea, the `encodeURIComponent` from https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/encode_overlay.js#L54 certainly appears to be suspicious.

scothis:
Closed via #156

PR #156, opened 2017-04-13
Avoid double URL encoding of geojson markers

scothis:
The previous geojson overlay was URL encoded internally and the value
was then url encoded again when it was placed into the URL. The value
should only be URL encoded when being placed into the URL.

This change also move the responsibility for the slash into the URI
template.

Issue: #154

### DIFF
--- a/lib/encode_overlay.js
+++ b/lib/encode_overlay.js
@@ -51,7 +51,7 @@ module.exports.encodePath = encodePath;
  * @private
  */
 function encodeGeoJSON(geojson) {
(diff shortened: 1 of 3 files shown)


scothis:
@andrewharvey can you give this branch a spin?

andrewharvey:
@scothis Looks good. I was trying to get it to encode other chars like `[]:` as if anyone then uses this URL in curl, they need to use `--globoff` to avoid curl reading a [0,0] as a glob, but I think this will be good enough as the primary use of this sdk is browser or node usage.

Thanks.

Issue #137, opened 2017-03-14
Make pagination easy


mcwhittemore:
For our endpoints that support pagination, we have a `next` header. This header can be added to `start` in the pagination options to go to the next page.

It would be cool if callbacks called by listing endpoints used a `function(err, data, next) {}` signature where next is a function that takes a callback only. When there are no more pages, next should be undefined.

tmcw:
I implemented multi-page support with Link headers in https://github.com/tmcw/got-links , could use that as an example.

andrewharvey:
How about treating this transparently to the user? The Styles API seems to limit results to 12 per page, if `client.listStyles`simply return all styles it would be much simpler.

andrewharvey:
@scothis Do you have any advice on the best way to do this, perhaps in the context of `rest`? I noticed the HATEOAS interceptor does very similar things but I couldn't manage to get it working.

https://github.com/mapbox/mapbox-sdk-js/commit/f55debb555beeef8d143b5a50fce393966b7dc43 is where I'm up to but it's not working. Any help would be much appreciated.

I know for dataset features endpoints, transparent pagination might not be best, but for things like styles or tilesets it makes things a lot easier on the client side.

mcwhittemore:
> The Styles API seems to limit results to 12 per page

Just FYI, this limit is based on the size of the style. If you had a bunch of really small styles you'd get more and if you had many very big style you'd get less. 

> How about treating this transparently to the user?

I'd be down for providing an opt-in option for requests. Something like `{getAll: true}` but it should be off for all endpoints by default. We don't want a situation where some endpoints work one way and other work another way.

scothis:
@andrewharvey rest.js has the facilities to parse out the Link header for pagination. The main issue is that we don't expose this capability in clean way. The question becomes how to expose pagination.

I'm a bit uneasy about offering an automatic tool to fetch all pages. These tools are helpful in the optimistic case, but will cause significant trouble when the user's expectations don't quite align. If a user really wants to do this, they can implement it themselves.

We can't implement the original suggestion of `(err, data, next) => {}` as the callback signature because the current signature is `(err, data, res) => {}`. Adding yet another arg is doable, but doesn't do anything to help users using Promises instead of callbacks.

I think the right way to handle this is to include a `.next([callback])` or `.nextPage([callback])` method on the response object when there is another page available. rest.js has a more generic notion of a `.follow(relationship)` method.

I'll take a crack at a quick PR.

[...]

PR #159, opened 2017-04-14
Response list pagination

scothis:
If there are more pages in a listing response, the response object
will now contain a `.nextPage()` method. This method will only be
available when there is another page of results available. Like the
original method, the `.nextPage()` method either accepts a callback
or returns a Promise. No other arguments are permitted.

Issue: #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,15 +44,41 @@ $ npm install --save mapbox
 
 ## Usage
 
-Basic usage of the geocoder:
+Setup:
 
 ```js
 var MapboxClient = require('mapbox');
 var client = new MapboxClient('YOUR_ACCESS_TOKEN');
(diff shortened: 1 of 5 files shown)


scothis:
In my local testing this code works. Creating automated tests will be a bit fun as we'll need to make sure there are enough resources to produce a paged result. This requires either assuming the account is pre populated with records, or seeding an account and then cleaning up without hitting a rate limit.

cc @mcwhittemore @andrewharvey

mcwhittemore:
@scothis some endpoints have a limit option, that should help for testing.

andrewharvey:
Thanks for working on his @scothis! I've added some comments and questions about the approach in #137.

Issue #157, opened 2017-04-13
Move URI template constants into services


scothis:
The idea around having the URI template constants be centralized is that it makes refactoring the URL space easier. This hasn't played out in practice, when an API version change does happen, it will still be limited to a single service.

As there is a tight coupling between the template and the code that fills the templated values, the URI templates should be moved from the constants file into the individual services so they are closer to the code they relate to.

andrewharvey:
closed via 5acd36b52ea1fdac35eb2a060794063364f6ada0 #160

PR #155, opened 2017-04-13
add tokens api support

andrewharvey:
Closes #143

~~I'd still like to do some more testing to and try to add more unit tests to ensure it's rock solid before merging.~~

PR #152, opened 2017-04-13
upgrade static service to include new static api and move old to static classic

andrewharvey:
Fixes #139

Implements leave the MapboxStatic service but rename the getStaticURL method to getStaticClassicURL and drop the new Static API into getStaticURL.

This is an API breaking change.

*moved from #149 for CI*

PR #160, opened 2017-04-15
Move URI template from constants into service

scothis:
The URI templates are tightly coupled to the parameters the template
accepts. The theory for centralizing all the templates into the
constants.js file was that it would make refactoring easier. It's very
unlikely that we will ever refactor the URL space where we would update
the space for more than one service.

Issue: #157


andrewharvey:
just a noting that someone will need to rebase depending on which gets merged first this or #155 / #152 as they have conflicting changes.

PR #166, opened 2017-05-08
post method for credentials

mapsam:
Using the POST endpoint for credentials instead of GET

cc @perrygeo @GretaCB


mapsam:
Thanks @scothis!

PR #171, opened 2017-06-02
fixes console.log

geografa:
results.origin was failing


geografa:
Hmm...There are a number of failing tests for individual services but not for this change. @mollymerp I will try to dig into the tests a little more next week.

geografa:
over here now: https://github.com/mapbox/mapbox-sdk-js/pull/172

Issue #146, opened 2017-04-12
improve documentation and link to Mapbox API docs


andrewharvey:
- [x] Taking advantage of a newer documentation.js version, API.md can be improved to include a TOC and more complete docs for all services and methods.

- [ ] The README.md should link to [`https://www.mapbox.com/api-documentation/?language=JavaScript`](https://www.mapbox.com/api-documentation/?language=JavaScript) which provides inline mapbox-sdk-js example usage.

/cc @lyzidiamond

andrewharvey:
I've attempted (1) in https://github.com/mapbox/mapbox-sdk-js/commit/d10eed9cfe4b09b62783937725eddbeeaeb8fe2c but it's not working that great yet.

itaditya:
Could you also mention all the api functions in the [doc](https://github.com/mapbox/mapbox-sdk-js/blob/master/API.md) first as a list (on top) so that we can know what functions exist, rather than to scroll all the time

andrewharvey:
> Could you also mention all the api functions in the doc first as a list (on top) so that we can know what functions exist, rather than to scroll all the time

@itaditya That's the intention of the .documentation.yml toc entry at https://github.com/mapbox/mapbox-sdk-js/commit/d10eed9cfe4b09b62783937725eddbeeaeb8fe2c but it needs some more work to fix outstanding issues.

mollymerp:
Thanks for picking this up @andrewharvey! I think a TOC would be awesome, and you're right about linking to the api-docs!! 🙇‍♀️ 
Let me know if you need anything to get your branch ready for review!

andrewharvey:
@mollymerp I don't have much time for this right now, I believe @geografa is working on this at #172. There is an open question about whether it's best to keep both API.md and the [api-documentation](https://www.mapbox.com/api-documentation/?language=JavaScript).

I find API.md is more complete and thorough than api-documentation but lacks the context of the APIs given at api-documentation. I'm leaning towards instead of trying to maintain two sets of docs, we settle on one. GL JS doesn't have an API.md instead using the Mapbox hosted docs, so this repo could do similar. :thinking:

PR #172, opened 2017-06-05
Update links and examples

geografa:
Starting this as a larger cleanup of docs and the API examples.


andrewharvey:
Isn't API.md auto generated by https://github.com/mapbox/mapbox-sdk-js/blob/master/package.json#L13, so it's better to fix the source and then rerun `npm run docs`.

andrewharvey:
See also #146 and https://github.com/mapbox/mapbox-sdk-js/commit/d10eed9cfe4b09b62783937725eddbeeaeb8fe2c

mollymerp:
These tests are flappy! I restarted and seems to be passing now.

geografa:
@scothis @mollymerp I was going to update services and change distance.js to matrix.js https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/services/distance.js

Ok to do this here or should I make a new PR?

README.md
```diff
@@ -31,14 +31,12 @@ Generally Available
 
 Contact help@mapbox.com for information
 
-* [Distance](https://www.mapbox.com/developers/api/distance/)
-  * Travel-time tables between up to 100 points
+* [Matrix](https://www.mapbox.com/api-documentation/#matrix)
```
mollymerp: I think this change is jumping the gun on #176 – once you revert this I think this PR will be good to go.

geografa: #176 has been merged into this so we should be gtg here

PR #177, opened 2017-06-09
updates API doc w/ getMatrix function

geografa:


API.md
```diff
@@ -1156,22 +1156,26 @@ for more documentation.
...
 var mapboxClient = new MapboxClient('ACCESSTOKEN');
 // With options
-mapboxClient.getDistances([
-  [-95.4431142, 33.6875431],
-  [-95.0431142, 33.6875431],
-  [-95.0431142, 33.0875431],
-  [-95.0431142, 33.0175431],
-  [-95.4831142, 33.6875431]
-], {
-  profile: 'walking'
```
danpat: Hmm, maybe the example should include the `profile` option, it's always nice to see explicit use in docs.

API.md
```diff
@@ -1156,22 +1156,26 @@ for more documentation.
         automobile and will use highways, `'walking'`, which avoids
         streets without sidewalks, and `'cycling'`, which prefers streets
         with bicycle lanes and lower speed limits for transportation via
-        bicycle. (optional, default `driving`)
+        bicycle. The `'mapbox/driving-traffic'` profile is not supported. (optional, default `driving`)
```
danpat: This should just say `driving-traffic` to be consistent with how the other profiles are specified.

The HTTP API uses the full `mapbox/driving-traffic` or `mapbox/driving` as the profile name, but it looks like the SDK automatically pre-pends `mapbox/`, so we probably shouldn't mention it.

API.md
```diff
@@ -1137,17 +1137,17 @@ mapboxClient.getDirections([
...
+Consult the [Mapbox Matrix API](https://www.mapbox.com/api-documentation/#matrix)
 for more documentation.
 
 **Parameters**
 
--   `waypoints` **Array&lt;Array&lt;number&gt;&gt;** an array of coordinate pairs
+-   `waypoints` **Object&lt;Object&lt;number&gt;&gt;** an array of coordinate object pairs
     in [longitude, latitude] order. Up to
-    100 waypoints can be specified.
+    25 waypoints can be specified.
```
danpat: Hmm, to avoid this number going stale, should we just refer to the online docs for "coordinate limits" ?  We may change/increase those at any time, it'd be nice not to need doc updates here when/if there are limit changes.

PR #188, opened 2017-08-30
Add dependencies used in Paginator

lesserj:
The paginator.js file has dependencies to "querystring" and "url" that are not resolved in the package.json file.  This caused a problem for me in a new project where the url library hadn't already been included by another package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,8 @@
   },
   "dependencies": {
     "es6-promise": "^4.0.5",
-    "rest": "^2.0.0"
+    "querystring": "^0.2.0",
+    "rest": "^2.0.0",
+    "url": "^0.11.0"
   }
 }


andrewharvey:
These modules are built into node https://nodejs.org/api/querystring.html https://nodejs.org/api/url.html, so when run in node it should work, and when run in the browser doesn't `browserify` take care of providing url and querystring?

I haven't run into this issue on node or in the browser, could you provide any more details which may help work out why you're running into an issue?

I don't think we should use the npm packages over the native node libraries.

lesserj:
Sorry, didn't see your reply.  I'm using this in a react-native environment which does not include those node modules.  They need to be manually added to the project.  See https://github.com/mapbox/mapbox-sdk-js/pull/196

andrewharvey:
> Sorry, didn't see your reply. I'm using this in a react-native environment which does not include those node modules. They need to be manually added to the project. See #196

I see, sorry I'm not sure myself so I'll leave this PR for Mapbox.

lesserj:
Thanks.  I think adding a note in the documentation would be satisfactory.

davidtheclark:
This is no longer an issue with the new codebase for `@mapbox/mapbox-sdk`. Please feel free to open a new issue if you have any similar troubles with the new package.

PR #196, opened 2017-11-02
Update README to show React Native install steps

nitaliano:
Currently this module does not work out of the box with react-native due to it's use of the url module. The standard node modules do not ship with react-native so if we polyfill the url module it does work. Figured we should just add a note about that incase anyone wants to also use this on RN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,15 @@ Generally Available
 $ npm install --save mapbox
 ```
 
+## React Native Installation
+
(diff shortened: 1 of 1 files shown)


davidtheclark:
@nitaliano Would you mind either refreshing this PR or opening a new one that targets the new codebase?

nitaliano:
Yup, I'll open a new one

PR #197, opened 2017-11-15
Fix #164: Allow query to be the only arg specified

### LINKED ISSUE #164: Geocoding Promise Syntax Inconsistency
Using the geocoding promise-based alternative syntax like the example in the README (without passing a options object) throws a `Uncaught Error: options must be an object`.

Is this the expected behavior? The example kind of implies that the options param is optional but reading the source expects one to be passed in when using the promise signature.

I prepared a Webpackbin to replicate the issue:
https://www.webpackbin.com/bins/-KjZU5GvhI1sBWs5NhAL

joshhendo:
The docs indicate that `options` is an optional parameter. At the moment, that's true if you're using the callback syntax, however if you're using it as a promise (and therefore only have one argument specified, `query` as a string), it throws an error.

Instead, we should allow options to be either a function or undefined. If `callback` and `options` are both undefined, I've left the existing code to work as is, so `callback` to `options` (undefined to undefined) and default options to an empty object. The alternative was to have multiple or nested if statements.

Unit tests may need to be updated - I had trouble running them (likely as I only had a read-only token at the time)

### DIFF
--- a/lib/services/geocoding.js
+++ b/lib/services/geocoding.js
@@ -70,9 +70,8 @@ function roundTo(value, places) {
  * });
  */
 MapboxGeocoding.prototype.geocodeForward = function(query, options, callback) {
-
-  // permit the options argument to be omitted
-  if (callback === undefined && typeof options === 'function') {
+  // permit the options argument to be omitted, or the options + callback args to be omitted if using promise syntax
+  if (callback === undefined && (options === undefined || typeof options === 'function')) {
     callback = options;
     options = {};
   }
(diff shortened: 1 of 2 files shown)


andrewharvey:
A unit test for this which fails on master, but passes on this branch would be good!

joshhendo:
Yep, I'll do that shortly. I've never used `tap` before so just got to get familiar with it first.

joshhendo:
Made some tests with `doesNotThrow` and passing in of only one (valid) argument.

joshhendo:
I agree, it should be a consistent experience across all the functions. If I get time I should be able to apply the same change to all of them in another PR.

Another thing I was thinking, but didn't do it to keep things consistent, is simply check for a truthy value instead of specifically checking for `undefined`. A rule of thumb that I've always held to is, unless there's a good reason to distinguish between `undefined` and `null` (and there sometimes is!) the check should simply be a truthy check... the main place this doesn't work is where it's a string of number and an empty string or 0 are valid values.

andrewharvey:
Thanks @joshhendo I've merged this one as it looks good to me. The other improvements can come as separate PRs.

andrewharvey:
My bad, I shouldn't have merged this yet, as the test case for reverse is broken. I didn't pick it up because the tests can only pass successfully in the Mapbox repo since they use a secret Mapbox token.

PR #198, opened 2017-11-18
include instructions for using client via a script tag

### LINKED ISSUE #170: a bundled version of library that doesn't require npm
hi there,

Is there - or could there be - a bundled version of the library hosted somewhere that doesn't require npm, for those wanting to pull this straight into our page much like we do the `mapbox gl js` library?

I couldn't see a reference to it in the readme.

Thanks
Charlie

andrewharvey:
Closes #170.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,18 @@ $ npm install --save mapbox
 
 ## Usage
 
-Setup:
-
+Setup via `node`:
 ```js
 var MapboxClient = require('mapbox');
+```
+
+Setup via `script` tag:
+```
+<script src='https://unpkg.com/mapbox@1.0.0-beta9/dist/mapbox-sdk.min.js'></script>
+```
+
+Initializing:
+```js
 var client = new MapboxClient('YOUR_ACCESS_TOKEN');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,10 +43,18 @@ $ npm install --save mapbox
 
 ## Usage
 
-Setup:
-
+Setup via `node`:
 ```js
```
stevage: You probably want to move the "Installation" section with `npm install --save mapbox` to here.

andrewharvey: I've added a commit moving the script tag usage up into installation. But open to any suggestions to make it clear and simple for both the browser and node.

andrewharvey:
I'm still not happy with these instructions, they feel jumbled. Probably the better solution is to include a full HTML example for use in the browser and keep the README focused at usage in node. But this will have to do until then.

Note to self. "Rebase and merge" won't add the PR number in parenthesis in the commit message but GitHub still shows the PR it come from when you click through to the commit.

"Squash and merge" does add the PR number in the commit message.

PR #201, opened 2017-12-06
fix tests for geocoding promise syntax from #197

andrewharvey:
There was an error in #197 which broke the unit tests. I didn't pick it up because the tests can only pass successfully in the Mapbox repo since they use a secret Mapbox token. :thinking:


andrewharvey:
This looks fixed now, the tests still failing seem unrelated.

PR #203, opened 2018-01-17
Requires an object

phocks:
Due to this line `invariant(typeof options === 'object', 'options must be an object');` https://github.com/mapbox/mapbox-sdk-js/blob/b68ef79af3123a89f4a6bbf3623c8777872a6508/lib/services/geocoding.js#L202 this seems to require an object otherwise you get an "Invariant Violation: options must be an object" error. An empty object will do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ client.geocodeForward('Chester, NJ', function(err, data, res) {
 As an alternative to callbacks, each method also returns a Promise:
 
 ```js
-client.geocodeForward('Chester, NJ')
+client.geocodeForward('Chester, NJ', {})
   .then(function(res) {
     // res is the http response, including: status, headers and entity properties
     var data = res.entity; // data is the geocoding result as parsed JSON


andrewharvey:
I think the underlying issue was actually fixed in https://github.com/mapbox/mapbox-sdk-js/pull/197 could you test against master to ensure this isn't necessary.

We should do a new release to push this fix out.

phocks:
Still getting

`Uncaught (in promise) Error: options must be an object at invariant`

seems to be at this line? https://github.com/mapbox/mapbox-sdk-js/blob/b68ef79af3123a89f4a6bbf3623c8777872a6508/lib/services/geocoding.js#L88

andrewharvey:
In this case though options should be getting set to the default `{}` at https://github.com/mapbox/mapbox-sdk-js/blob/b68ef79af3123a89f4a6bbf3623c8777872a6508/lib/services/geocoding.js#L74 :confused:

phocks:
Hmm strange. Oh well it works for me just putting the {} in manually anyway. Thanks, maybe wait and see if any others have the same issue. Maybe it's just me.

sethomas:
+1, also getting this error, I manually put in `{}` as the second arg next to query and all is well

PR #205, opened 2018-03-04
Add options parameter to readme examples.

econner:
Using the promise example as written in README.md leads to an error:

> invariant.js:42 Uncaught (in promise) Error: options must be an object

The callback version works I guess because the callback overloads that parameter? Not really sure.

Anyway I added these arguments so the examples work as written if you just copy and paste code off internet READMEs like I do :-).

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ var client = new MapboxClient('YOUR_ACCESS_TOKEN');
 Basic usage of the geocoder:
 
 ```js
-client.geocodeForward('Chester, NJ', function(err, data, res) {
+client.geocodeForward('Chester, NJ', { /* optional params */ },  function(err, data, res) {
   // data is the geocoding result as parsed JSON
   // res is the http response, including: status, headers and entity properties
 });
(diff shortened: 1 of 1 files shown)


econner:
Well that method clearly takes an options parameter, so I'm not really sure how it can be fixed given the syntax currently in the readme unless you actually want to allow overloading the options parameter with a callback? 

```
MapboxGeocoding.prototype.geocodeForward = function(query, options, callback) {
// ...
}
```

That seems like a bad idea to me that will lead to confusion by users of this library, but I leave it up to you. In my opinion it's more clear this way that the method actually takes options... A user might never discover that if you allow the callback to be overloaded.

andrewharvey:
Sorry I just realised a release was never done after https://github.com/mapbox/mapbox-sdk-js/pull/197 was merged.

Could you try updating to `master` it should fix the issue for you? I'll try to coordinate a new release but will need Mapbox to actually publish it.

> ... unless you actually want to allow overloading the options parameter with a callback?

That's the intention of the current code https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/services/geocoding.js#L72 if options is a function it will use that as the callback.

PR #209, opened 2018-03-20
add table of contents and groups to API.md

andrewharvey:
open to suggestions if there is a better way to do this, but this is the minimal change I could do to get the desired result


jfirebaugh:
documentation.js can infer the name and membership from a construct like `Datasets.prototype.listDatasets = ...`, but it needs the parent `Dataset` to have a JSDoc comment as well. You can use the `documentation lint` command to catch cases where this is missing. E.g. if I revert the `@name` additions in `services/datasets.js`, `documentation lint` says:

```
   14:1  warning  @memberof reference to Datasets not found
   64:1  warning  @memberof reference to Datasets not found
  107:1  warning  @memberof reference to Datasets not found
  142:1  warning  @memberof reference to Datasets not found
  185:1  warning  @memberof reference to Datasets not found
  213:1  warning  @memberof reference to Datasets not found
  273:1  warning  @memberof reference to Datasets not found
  285:1  warning  type object found, Object is standard
  366:1  warning  @memberof reference to Datasets not found
  407:1  warning  @memberof reference to Datasets not found
```

lib/services/datasets.js
```diff
@@ -4,7 +4,9 @@ var invariant = require('../../vendor/invariant');
 var hat = require('../../vendor/hat');
 var makeService = require('../make_service');
 
-var Datasets = module.exports = makeService('MapboxDatasets');
+/**
+ */
```
mollymerp: maybe add a `@class` tag so its clear in the future why these empty comment blocks are here 🙃

andrewharvey:
Thanks for the great feedback @mollymerp @jfirebaugh, I've folded in the changes you've suggested.

PR #218, opened 2018-05-09
Create service clients without first creating a base client

davidtheclark:
- Introduces the ability to create service clients without first creating a base client. For many use cases this will clarify things, cut down on an abstract step. Hopefully the documentation changes in the README explain this new power.
- Adds `npm run try-browser`, which uses budo to open a simple page set up for manual testing, similar to `test/try-node.js` for Node. I wanted this as a sanity check that my change worked in the browser. It's very simple but does the trick, for sanity checking.

@kepta for review.

@emilymdubois also interested in any feedback you might have, since this is a slight change (addition) to the public API for the package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,17 @@ There are 3 basic steps to getting an API response:
 
 ### Creating clients
 
-To **create a base client**, import the factory function from `'@mapbox/mapbox-sdk'` and provide it with your access token.
+To **create a service client**, import the service's factory function from `'@mapbox/mapbox-sdk/services/{service}'` and provide it with your access token.
 
-The then **create a service client**, import the service's factory function from `'@mapbox/mapbox-sdk/services/{service}'` and provide it with your base client.
+The service client exposes methods that create requests.
+
+```js
+const mbxStyles = require('@mapbox/mapbox-sdk/services/styles');
+const stylesService = mbxStyles({ accessToken: MY_ACCESS_TOKEN });
+// stylesService exposes listStyles(), createStyle(), getStyle(), etc.
+```
+
+To **share configuration between multiple services**, you can create a base client and then pass *that* into the service factory functions.
 
 ```js
 const mbxClient = require('@mapbox/mapbox-sdk');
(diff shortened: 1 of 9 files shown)


emilymdubois:
@davidtheclark this seems like a lovely change. Only suggestion I have is to tweak the language in the README to clarify a client may be used to share configurations among multiple service factory functions, and there are two ways to accomplish that (example below).

I'm really enjoying this browser testing tool by the way! I can see this being extremely useful for testing API methods that are utilized in front-end applications but are difficult to reproduce in the UI, and for generating test fixtures.

---

You may use a client to **share configuration between multiple services**. There are two ways to do this.

1. Initialize a base client and then pass *that* into the service factory functions.

```js
const mbxClient = require('@mapbox/mapbox-sdk');
const mbxStyles = require('@mapbox/mapbox-sdk/services/styles');
const mbxTilesets = require('@mapbox/mapbox-sdk/services/tilesets');

const baseClient = mbxClient({ accessToken: MY_ACCESS_TOKEN });
const stylesService = mbxStyles(baseClient);
const tilesetsService = mbxTilesets(baseClient);
```

2. Initialize a service, and pass the base client exposed on its `client` property into additional service factory functions.

```js
const mbxStyles = require('@mapbox/mapbox-sdk/services/styles');
const mbxTilesets = require('@mapbox/mapbox-sdk/services/tilesets');

const stylesService = mbxStyles({ accessToken: MY_ACCESS_TOKEN });
const tilesetsService = mbxTilesets(stylesService.client);
```

kepta:
I feel we should not document the other approach
```javascript
const stylesService = mbxStyles({ accessToken: MY_ACCESS_TOKEN });
const tilesetsService = mbxTilesets(stylesService.client);
```
because 
1. I don't think it adds anything more than what the previous approach gives. Let us let the user worry on how to import `baseClient` rather than giving them a way to extract it from services. 
2. Increases the surface area of the public API, adds trouble for the maintainer.

I feel being a bit opinionated and focusing on providing a single instead of multiple way of doing a thing would help us in the long run.

davidtheclark:
@emilymdubois @kepta Thanks for the feedback. I agree with both of you — so I'll change the docs accordingly.

PR #217, opened 2018-05-07
Revitalization

davidtheclark:
Addresses #211.

This branch can serve as the base branch for revitalization work. In this initial commit, I've transferred code over from our internal prototype. Follow-ups can be PRs into this branch.

The diff isn't going to be useful, because the codebase is completely replaced. So please pull down the branch and look around, and share your feedback!

A few details about the codebase:

- All the syntax is ES5-compatible, so we don't have to worry about a compilation step (which is especially nice in early stages of development).
- MD files in `docs/` are generated by documentation.js from code comments.
- Unit tests are in `__tests__/` directories next to the files they test. Integration tests are in `test/`. `test-shared-interface.js` includes integration tests that are run against both Node and browser clients.
- There's a little CLI in `test/try-node.js` that we can use to check responses when adding new services and service methods. We could also build something for the browser, but that might be more work.
- The `"browser"` field in `package.json` should instruct bundlers targeting the browser to use the browser client instead of the Node client. (I've tried this out, and seems to work fine.)
- Prettier formats the JS, remark formats the MD.

(@kepta @emilymdubois — FYI: In creating this branch, I've made some changes to directory structure to take advantage of the new setting. I hope it's a little better organized now. Also, I refactored the validators some, hoping to make its usage a little more straightforward.)

**TODO**

- [x] Add services to cover everything that is already covered by the public SDK. (@tristen were you interested in helping out with this step?). Below is a checklist for the existing SDK methods that we'll need equivalents for
  - [x] MapboxStyles @davidtheclark
      - [x] listStyles
      - [x] createStyle
      - [x] updateStyle
      - [x] deleteStyle
      - [x] readStyle
      - [x] readSprite
      - [x] readFontGlyphRanges
      - [x] addIcon
      - [x] deleteIcon
      - [x] embedStyle
  - [x] MapboxStatic @davidtheclark
      - [x] getStaticURL
      - ~[ ] getStaticClassicURL~
  - [x] MapboxUploads @kepta
      - [x] listUploads
      - [x] createUploadCredentials
      - [x] createUpload
      - [x] readUpload
      - [x] deleteUpload
  - [x] MapboxDatasets @davidtheclark
      - [x] listDatasets
      - [x] createDataset
      - [x] readDataset
      - [x] updateDataset
      - [x] deleteDataset
      - [x] listFeatures
      - [x] insertFeature
      - [x] readFeature
      - [x] deleteFeature
  - [x] MapboxTilesets @davidtheclark
      - [x] tilequery
[...]

PR #219, opened 2018-05-13
[REVITALIZATION] Fix error on multiple simultaneous requests; clarify docs

davidtheclark:
@kepta: This address your comment here: https://github.com/mapbox/mapbox-sdk-js/pull/217#issuecomment-388317162

It was already behaving as expected if you re-sent after a response or error were received (and this was tested). It was *not* behaving the same way if you sent multiple times right away. That was a bug, so I fixed it, added a test, and then added some clarifying docs.

> How does .eachPage signal done?

This is like `forEach`: the callback just doesn't run if there's not another page to run it on. Does that make sense? Is there something you think we need to say to clarify this?

### DIFF
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -101,6 +101,10 @@ The `emitter` property is an `EventEmitter` that emits the following events:
     received an error in response, the error is available on this property.
 - `aborted` **[boolean][15]** If the request has been aborted
     (via [`abort`][8]), this property will be `true`.
+- `sent` **[boolean][15]** If the request has been sent, this property will
+    be `true`. You cannot send the same request twice, so if you need to create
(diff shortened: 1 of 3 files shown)


kepta:
> This is like forEach: the callback just doesn't run if there's not another page to run it on. Does that make sense? Is there something you think we need to say to clarify this?

This makes total sense, It just occurred to me that the `.next` style is quite a lot similar to the [iteration protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) of javascript and also the [subscribe](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/subscribe.md) of rxjs, so thats why I brought the `.done/.complete`.

PR #221, opened 2018-05-14
[REVITALIZATION] Test progress events

kepta:
This pull request aims to add progress events under test coverage.


davidtheclark:
@kepta Yeah, we might just have to live with those differences — in which case we'll want to implement separate tests for each environment. 

For the browser, I think using XHR's native events is fine, even if they sometimes behave kind of strangely. Unless you can figure out a better way that is also relatively simple?

kepta:
@davidtheclark I have segregated the node and browser testing and I tried using the `xhr-mock` library for the browser tests.

lib/classes/mapi-request.js
```diff
@@ -4,7 +4,7 @@ var parseToken = require('@mapbox/parse-mapbox-token');
 var xtend = require('xtend');
 var EventEmitter = require('eventemitter3');
 var urlUtils = require('../helpers/url-utils');
-
+var constants = require('../constants');
 var requestId = 1;
```
davidtheclark: I'd prefer to conventionally keep an empty line separating imports from the rest of the code.

test/browser-interface.test.js
```diff
@@ -19,3 +20,59 @@ test('errors early if access token not provided', () => {
     }
   );
 });
+
+describe('test node progress events', () => {
```
davidtheclark: node > browser

test/browser-interface.test.js
```diff
@@ -19,3 +20,59 @@ test('errors early if access token not provided', () => {
...
+
+  beforeEach(() => {
+    mockXHR.setup();
+    const accessToken = mockToken();
+    mockXHR.get(
+      `https://api.mapbox.com/styles/v1/mockuser/foo?access_token=${accessToken}`,
+      {
+        status: 200,
+        headers: {
+          'Content-Length': '18',
```
davidtheclark: Would you mind adding a comment with the URL of the mockXhr docs that explain why this works? I expect I'd be confused by this if I look at it again 4 months later.

test/browser-interface.test.js
```diff
@@ -19,3 +20,59 @@ test('errors early if access token not provided', () => {
...
+    const client = browserClient({ accessToken });
+
+    request = client.createRequest({
+      method: 'GET',
+      path: '/styles/v1/:ownerId/:styleId',
+      params: { styleId: 'foo' }
+    });
+  });
+
+  test('request.emitter should emit uploadProgress events', () => {
```
davidtheclark: This should say "should *not* emit", right?

kepta:
@davidtheclark improved the code, let me know what you think.

PR #224, opened 2018-05-19
Add Style#getStyleSprite, getFontGlyphRange, getEmbeddableHtml

davidtheclark:
Style#getStyleSprite replaces getStyleSpriteJson to allow for PNGs, also.

Removes the requirement that the response's body be parseable as JSON, because these endpoints receive non-JSON.

@kepta for review.

cc @emilymdubois

### DIFF
--- a/services/styles.js
+++ b/services/styles.js
@@ -254,7 +254,7 @@ Styles.getStyleSprite = function(config) {
  * See [the public documentation](https://www.mapbox.com/api-documentation/?language=JavaScript#retrieve-font-glyph-ranges).
  *
  * @param {Object} config
- * @param {Array<string>} config.fonts - An array of font names.
+ * @param {string|Array<string>} config.fonts - An array of font names.
  * @param {number} config.start - Character code of the starting glyph.
  * @param {number} config.end - Character code of the last glyph,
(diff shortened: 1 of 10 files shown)


services/styles.js
```diff
@@ -215,34 +216,120 @@ Styles.deleteStyleIcon = function(config) {
...
+      scrollZoom: v.boolean,
+      title: v.boolean,
+      ownerId: v.string
+    },
+    config
+  );
+
+  var scrollZoom = config.scrollZoom !== undefined ? config.scrollZoom : true;
+  var title = config.title !== undefined ? config.title : false;
+  var fileName = config.styleId + '.html';
```
kepta: nit: I guess the rest endpoint would automatically assume the defaults. Do you think it makes sense to send redundant data?

davidtheclark: > Do you think it makes sense to send redundant data?

Yeah, I'll cut that.

davidtheclark:
Thanks, @kepta. I'm going to wait to finish and merge this until https://github.com/mapbox/mapbox-sdk-js/pull/226, because that will allow me to cut a couple of lines.

PR #227, opened 2018-05-21
Directions

kepta:
This adds `getDirections`.

In this PR I ended up adding a bunch of extra validators. Seeing them now I wonder if I just made them a bit complex and unreadable.
I also tried to document the parameters looking at [API docs](https://www.mapbox.com/api-documentation/#retrieve-directions), but then I thought it might be a futile approach to keep the docs in sync at two places. I noticed the earlier implementation had pretty dense documentation but most of it was out of sync with the REST api.
thoughts @davidtheclark ?

Pending actions
- [x] Test them manually using real endpoints.
- [x] 👍  get a thumbs or feedback with the overall strategy of tacking this endpoint.

Side Note: I have no idea on how to avoid restricting the markdown files. I removed `node_modules`, did `npm install`, even tried `npm ci`, but the problem doesn't seem to go away.

### DIFF
--- a/services/directions.js
+++ b/services/directions.js
@@ -16,7 +16,7 @@ var Directions = {};
  *
  * @param {Object} config
  * @param {'driving-traffic'|'driving'|'walking'|'cycling'} [config.profile="driving"]
- * @param {Array<wayPoints>} config.wayPoints - An ordered array of `wayPoint` object. There can be between 2 and 25 wayPoints.
+ * @param {Array<waypoints>} config.waypoints - An ordered array of `waypoint` object. There can be between 2 and 25 waypoints.
  * @param {boolean} [config.alternatives=false] - Whether to try to return alternative routes.
  * @param {Array<'duration'|'distance'|'speed'|'congestion'>} [config.annotations] - Whether or not to return additional metadata along the route.
  * @param {boolean} [config.bannerInstructions=false] -  Should be used in conjunction with `steps`.
@@ -35,7 +35,7 @@ Directions.getDirections = function(config) {
   v.validate(
     {
       profile: v.oneOf('driving-traffic', 'driving', 'walking', 'cycling'),
-      wayPoints: v.arrayOf(v.plainObject).required,
+      waypoints: v.arrayOf(v.plainObject).required,
       alternatives: v.boolean,
       annotations: v.arrayOf(
         v.oneOf('duration', 'distance', 'speed', 'congestion')
@@ -62,15 +62,15 @@ Directions.getDirections = function(config) {
    * all the properties that depend on the order of coordinates into
    * one object for ease of use.
    *
-   * @typedef {Object} wayPoints
+   * @typedef {Object} waypoints
    * @property {number} latitude
    * @property {number} longitude
    * @property {'unrestricted'|'curb'} [approach="unrestricted"] - Used to indicate how requested routes consider from which side of the road to approach a waypoint.
    * @property {Array<number>} [bearing] - Used to filter the road segment the waypoint will be placed on by direction and dictates the angle of approach.
    * @property {number|'unlimited'} [radius] - Maximum distance in meters that each coordinate is allowed to move when snapped to a nearby road segment.
-   * @property {string} [wayPointName] - Custom names for waypoints used for the arrival instruction in banners and voice instructions.
+   * @property {string} [waypointName] - Custom names for waypoints used for the arrival instruction in banners and voice instructions.
    */
-  config.wayPoints.forEach(function(wayPoint) {
+  config.waypoints.forEach(function(waypoint) {
     v.validate(
       {
         latitude: v.number.required,
@@ -78,49 +78,49 @@ Directions.getDirections = function(config) {
         approach: v.oneOf('unrestricted', 'curb'),
         bearing: v.arrayOf(v.number),
         radius: v.oneOfType(v.number, v.oneOf('unlimited')),
-        wayPointName: v.string
+        waypointName: v.string
       },
-      wayPoint
+      waypoint
     );
   });
 
-  var wayPoints = {
+  var waypoints = {
     coordinates: [],
     approach: [],
     bearing: [],
     radius: [],
-    wayPointName: []
+    waypointName: []
   };
 
-  config.wayPoints.forEach(function(wayPoint) {
-    wayPoints.coordinates.push(wayPoint.longitude + ',' + wayPoint.latitude);
+  config.waypoints.forEach(function(waypoint) {
+    waypoints.coordinates.push(waypoint.longitude + ',' + waypoint.latitude);
 
     // join props which come in pairs
     ['bearing'].forEach(function(prop) {
-      if (wayPoint.hasOwnProperty(prop) && wayPoint[prop] != null) {
-        wayPoint[prop] = wayPoint[prop].join(',');
+      if (waypoint.hasOwnProperty(prop) && waypoint[prop] != null) {
+        waypoint[prop] = waypoint[prop].join(',');
       }
     });
 
-    ['approach', 'bearing', 'radius', 'wayPointName'].forEach(function(prop) {
-      if (wayPoint.hasOwnProperty(prop) && wayPoint[prop] != null) {
-        wayPoints[prop].push(wayPoint[prop]);
+    ['approach', 'bearing', 'radius', 'waypointName'].forEach(function(prop) {
+      if (waypoint.hasOwnProperty(prop) && waypoint[prop] != null) {
+        waypoints[prop].push(waypoint[prop]);
       } else {
-        wayPoints[prop].push('');
+        waypoints[prop].push('');
       }
     });
   });
 
-  ['approach', 'bearing', 'radius', 'wayPointName'].forEach(function(prop) {
+  ['approach', 'bearing', 'radius', 'waypointName'].forEach(function(prop) {
     // avoid sending params which are all `;`
     if (
-      wayPoints[prop].every(function(char) {
+      waypoints[prop].every(function(char) {
         return char == '';
       })
     ) {
-      delete wayPoints[prop];
+      delete waypoints[prop];
     } else {
-      wayPoints[prop] = wayPoints[prop].join(';');
+      waypoints[prop] = waypoints[prop].join(';');
     }
   });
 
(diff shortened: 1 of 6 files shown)


services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+var Directions = {};
+
+/**
+ * Get directions.
+ *
+ * See the [Mapbox Directions API](https://www.mapbox.com/api-documentation/#directions).
+ *
+ * @param {Object} config
+ * @param {'driving-traffic'|'driving'|'walking'|'cycling'} [config.profile=driving]
```
davidtheclark: I think default values that are strings in JSDoc should have quotations around them? `[config.profile="driving"]`

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {Array.<Object>} config.way_points - An ordered array of objects with `latitude`
```
davidtheclark: Let's use camelCase to fit with the convention everywhere else in the library.

davidtheclark: Flow-type annotations should work, so `Array<Object>` (no need for the period). In fact, we may want to see what happens in the doc if we flesh out this annotation, e.g. `Array<{ longitude: number, latitude: number, approach?: 'unrestried' | 'curb', ...}>`

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ * and `longitude` properties that represent points to visit and optional `approach`, `bearing`, `radius` & `waypoint_name`,
+ * see the [Mapbox Directions API](https://www.mapbox.com/api-documentation/#retrieve-directions) for details.
+ * There can be between 2 and 25 coordinates.
+ * @param {string} [config.alternatives=false]
```
davidtheclark: Boolean.

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {Array.<string>} [config.annotations]
```
davidtheclark: Let's provide the possibilities.

services/directions.js
davidtheclark: I think it would be very helpful as a user to explain how I can look up the docs about each of the options in these objects. Maybe just a sentence about how these are listed as independent query strings whose length must match the waypoints' length, but they are combined here into one object for each waypoint.

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {boolean} [config.banner_instructions=false]
+ * @param {boolean} [config.continue_straight]
+ * @param {string} [config.exclude]
```
davidtheclark: For each of these with a finite set of options, let's specify the strings that are available (if the list is reasonably short).

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {string} [config.language=en]
+ * @param {string} [config.overview=simplified]
+ * @param {boolean} [config.roundabout_exits=false]
+ * @param {boolean} [config.steps=false]
+ * @param {boolean} [config.voice_instructions=false]
+ * @param {string} [config.voice_units=imperial]
+ * @return {MapiRequest}
+ */
+Directions.getDirections = function(config) {
+  config.profile = config.profile || 'driving';
```
davidtheclark: We've been validating first, setting defaults next.

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+    waypoint_name: []
+  };
+
+  config.way_points.forEach(function(wayPoint) {
+    wayPoints.coordinates.push(wayPoint.longitude + ',' + wayPoint.latitude);
+
+    // join props which come in pairs
+    ['bearing'].forEach(function(prop) {
+      if (wayPoint.hasOwnProperty(prop) && wayPoint[prop] != null) {
+        wayPoint[prop] = wayPoint[prop].join(',');
```
davidtheclark: After 843fc2693108a427f3cfb5b888ff51aa2c59de87, we'll want this to be an array instead of a string, so the comma isn't encoded.

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+  ['approach', 'bearing', 'radius', 'waypoint_name'].forEach(function(prop) {
+    // avoid sending params which are all `;`
+    if (
+      wayPoints[prop].every(function(char) {
+        return char == '';
+      })
+    ) {
+      delete wayPoints[prop];
+    } else {
+      wayPoints[prop] = wayPoints[prop].join(';');
```
davidtheclark: Do we need to ensure that this semicolon is not encoded, or is it ok?

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+    roundabout_exits: config.roundabout_exits,
+    steps: config.steps,
+    voice_instructions: config.voice_instructions,
+    voice_units: config.voice_units,
+
+    approach: wayPoints.approach,
+    bearing: wayPoints.bearing,
+    radius: wayPoints.radius,
+    waypoint_name: wayPoints.waypoint_name
+  };
```
davidtheclark: You could probably use `pick` and `xtend` utils for this, if you wanted. But it's fine as-is, too.

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+
+  // remove null/undefined keys
+  Object.keys(query).forEach(function(key) {
+    if (query[key] == null) {
+      delete query[key];
+    }
+  });
```
davidtheclark: Let's make a shareable utility function for this, in case it comes up again. Maybe something like `pickDefined`?

services/directions.js
```diff
@@ -0,0 +1,151 @@
...
+
+  return this.client.createRequest({
+    method: 'GET',
+    path: '/directions/v5/:profile/:coordinates',
+    params: {
+      profile: config.profile,
+      coordinates: wayPoints.coordinates.join(';')
```
davidtheclark: Again wondering if we'll have a problem with encoded semicolons.

services/service-helpers/validator.js
```diff
@@ -23,6 +23,73 @@ v.validate = function(schema, config) {
   });
 };
 
+v.oneOf = function(options) {
```
davidtheclark: Let's make this `oneOfType`, do distinguish from `oneOf` which compares against definite values. 

And what do you think of using `arguments` instead of an array argument? Curious so I can be consistent with the decision.

services/service-helpers/validator.js
```diff
@@ -23,6 +23,73 @@ v.validate = function(schema, config) {
...
+  if (!Array.isArray(options)) {
+    throw new Error('oneOf expects an array');
+  }
+  var checks = [];
+  options.forEach(function(option) {
+    if (!option || !option.__check) {
+      throw new Error(option + ' is not a valid option');
+    }
+    checks.push(option.__check);
+  });
```
davidtheclark: I'm not sure we should add code in here to validate the validation, since this doesn't help the end user, just the developer. With the ideal of making this validator as small as possible, could we could some lines and expect the developer to follow precedent and debug? (We could also document the validators.)

services/service-helpers/validator.js
```diff
@@ -23,6 +23,73 @@ v.validate = function(schema, config) {
...
+      .map(function(check) {
+        return check(value);
+      })
+      .filter(function(message) {
+        return message;
+      });
+
+    // no match
+    if (messages.length === checks.length) {
+      return messages.join(' or ');
```
davidtheclark: The resultant message is a little bit awkward: `{something} must be a string or {something} must be a number`. A more natural sentence would be `{something} must be a string or a number`. Is there a way we can rearrange where & how `must be` is inserted to allow for that?

services/service-helpers/validator.js
```diff
@@ -23,6 +23,73 @@ v.validate = function(schema, config) {
...
+      var message = values
+        .map(function(val) {
+          return check(val);
+        })
+        .find(function(message) {
+          return message;
+        });
+
+      if (message) {
+        return "array's every element " + message;
```
davidtheclark: How about `"must be an array whose every element" + message`?

services/service-helpers/validator.js
```diff
@@ -73,6 +140,17 @@ v.arrayOfStrings = wrapCheck(function(value) {
...
+v.arrayOfObjects = wrapCheck(function(value) {
+  if (
+    !Array.isArray(value) ||
+    !value.every(function(x) {
+      return typeof x === 'object';
+    })
+  ) {
+    return 'must be an array of objects';
+  }
+});
```
davidtheclark: Let's use the `arrayOf` validator you wrote, unless you want to back out of that. We can follow up on this PR with another that replaces `arrayOfStrings` and whatever related functions with `arrayOf`.

services/service-helpers/validator.js
```diff
@@ -114,6 +192,11 @@ function wrapCheck(check) {
     wrappedRequired(config, key);
     wrapped(config, key);
   };
+
+  wrapped.required.__required_check = check;
```
davidtheclark: A comment explaining this would be nice 👍

services/directions.js
kepta: I this both of them are fine [jsdoc](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values).

For consistency I will go ahead and quotes around it.

services/directions.js
kepta: The `{key?: value}`  doesn't seem to work with it, I thinking of using `typedef` to document `wayPoints`.

services/directions.js
kepta: I tried running this
```javascript
appendQueryObject('/foo/bar?one=1', {
        two: 2,
        three: '3,3;3,3'
})
```
Where `three` follows the the `way_points.coordinate` style of `,` between lat/lons and `;` between other coordinates. 

I get this output:
`/foo/bar?one=1&two=2&three=3%2C3%3B3%2C3`

Which seems fine to me, 🤷‍♂️ .

services/directions.js
kepta: I modified the `pick` and overloaded the second param to accept a callback which might help me in this case.

services/directions.js
kepta: As I was unable to add options properties, I have moved the entire `wayPoints` type into a typeDef.

services/service-helpers/validator.js
kepta: I am assuming that all messages must follow the convention `must be {XYZ}`. Which allows me to remove all of the `must be` to possibly fix the problem. Thoughts ?

davidtheclark: Yes, I think they should all be in the form of `must be {a|an} {thing}`.

kepta:
@davidtheclark this is ready for a review.
I tested it with the real api and played around with the query params, seems to work.

PR #228, opened 2018-05-22
Datasets and Tilequery services

davidtheclark:
Adds Datasets and Tilequery services.

@kepta for review.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -6,6 +6,7 @@
 - The following verbs should be used for each method type:
   - `get` and `list` for `GET` requests
   - `create` for `POST` requests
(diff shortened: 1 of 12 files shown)


davidtheclark:
Noting that `Datasets#updateFeature` is a little weird for naming: because it's a `PUT` request, you could create a new feature or update an existing one. The current language is `insertFeature`, but I don't really think that's any more clear. `putFeature` might be the most accurate, and clear for people who know about `PUT` requests. @emilymdubois do you think we should introduce `put` to the vocabulary for this?

kepta:
I like using `put` over `insert`, as it can upsert a feature. Also `put` is a well known HTTP verb.

PR #230, opened 2018-05-22
[WIP] Revamp validator

davidtheclark:
@kepta I took at a shot at a little proof of concept of the validator changes I had in mind. I'd like to turn this over to you from here. You could take or leave the ideas in here — or if you think that there's an existing library that would fit our usage, we could try that, too.

I think it may be important to sort this out before porting more services (so we have to change less validation later).

What do you think?


davidtheclark:
Another thing to consider — as we chatted about — would be to open source this as a separate library. The things that would distinguish it from other type assertion libraries I've been able to find, at least, are 

(a) it's browser-friendly (as small as possible, compiled to run in the browser, etc)
(b) its primary purpose is _not_ the type assertion itself, but (like `prop-types`) presenting an API for making assertions in a schema-like way, with nice error messages; it would provide minimal assertions out-of-the-box with the option to easily add custom assertions. (In our case, the existing `date` assertion would definitely be a custom one, specific to this SDK.)

Do you think that would be worth the trouble? Maybe a project we could save for a later date ...

kepta:
@davidtheclark this is fantastic!!
This would solve be super helpful as a lib.
I went ahead and added:
- couple of minor fixes.
- exposed a couple of internal functions so folks can create their own validators.

davidtheclark:
Great @kepta. Let's get this to the point that we can merge it into the SDK (and flesh out its unit testing), then we can look into extracting it after we get the SDK finished 👍

davidtheclark:
@kepta If you haven't yet, you might want to look into using [prop-types](https://github.com/facebook/prop-types). It is essentially the same API that we're going for, right?

Some pros for using prop-types (assuming it works):

- External dependency we don't have to test but know will work. Very thoroughly battle tested by React usage.
- Allows for all the use cases we need, including custom validators, without including a *ton* of stuff we don't, like Node libraries.

Some cons:

- It's not *super* small. (But I don't think it's *super* big, or inappropriately big. If we were to build out very similar functionality we'd probably end up with something the same size, right?)
- Its language is tied to React. Maybe we could use the arguments in https://github.com/facebook/prop-types#proptypescheckproptypes to tailor that, though.
- I wonder if we need to be concerned about how tightly it's tied up with React?

davidtheclark:
I found https://github.com/kentcdodds/api-check, which aims to be a prop-types-without-react.

kepta:
@davidtheclark I think we are better off building our own validation library, since it is central to a lot of our tools. 

I am on fence regarding `ketcdodds`'s `api-check`, since it does exactly what we need. The only concern is that it is a bit dated.

kepta:
@davidtheclark I played around trying to implement `shapeOf` & helpful path included in error messages. Tried to implement it with the `validator2`, but then ended up creating a `validator3` 😛. 

I tried to make it a bit versatile by allowing custom error messages and simpler validators. 

Let me know what you think. If you are 👍 with this approach I will go ahead and implement the remaining types and convert the existing code to use this.

services/service-helpers/validator3.js
```diff
@@ -0,0 +1,173 @@
...
+    var path = messages.slice(0, len - 1);
+
+    if (path.length == 0) {
+      // Calling it value since there is no identifiable path
+      path = ['value'];
+    }
+
+    var errorMessage;
+
+    if (typeof lastElement == 'function') {
```
davidtheclark: Why just two `==`?

services/service-helpers/validator3.js
```diff
@@ -0,0 +1,173 @@
...
+        typeof path[0] == 'number'
+      ) {
+        startWith = 'item at position ';
+      }
+
+      errorMessage = formatMessage(
+        apiName,
+        startWith + path.join('.') + ' must be ' + lastElement
+      );
+    }
```
davidtheclark: I think it would clarify the code to extract this if-else block into a separate function that returns `errorMessage`. You could use an early return and reduce some nesting.

services/service-helpers/validator3.js
```diff
@@ -0,0 +1,173 @@
...
+
+  var result = validator(value);
+
+  if (result) {
+    return Array.isArray(result) ? result : [result];
+  }
+}
+
+v.shapeOf = function shapeOf(validatorObj) {
+  var validatorZipped = objectEntries(validatorObj);
```
davidtheclark: `zip` is a little confusing to me here, because I don't think the transform corresponds to `_.zip`. Am I just missing the connection?

services/service-helpers/validator3.js
```diff
@@ -0,0 +1,173 @@
...
+  return function shapeOfValidator(value) {
+    var validationResult = validate(v.plainObject, value);
+
+    if (validationResult) {
+      return validationResult;
+    }
+
+    for (var i = 0; i < validatorZipped.length; i++) {
+      var key = validatorZipped[i].key;
+      var validator = validatorZipped[i].val;
```
davidtheclark: I think it's good practice with `var` to extract it from blocks that it's not actually scoped to, like for loops, to make it more clear what's going on. So 

```js
var i, key, validator;
for (i = 0; i < validatorZipped.length; i++) {
  key = ..
  validator = ..
}
```

services/service-helpers/validator3.js
```diff
@@ -0,0 +1,173 @@
...
+  var result = validator(value);
+
+  if (result) {
+    return Array.isArray(result) ? result : [result];
+  }
+}
+
+v.shapeOf = function shapeOf(validatorObj) {
+  var validatorZipped = objectEntries(validatorObj);
+  return function shapeOfValidator(value) {
```
davidtheclark: I think it'd be nice to have a comment describing the return value of this and similar functions. It's really cool how you worked this out.

services/service-helpers/__tests__/validator3.test.js
```diff
@@ -0,0 +1,246 @@
...
+  describe('nested require', () => {
+    var validator = v.warn(
+      v.shapeOf({ person: v.req(v.shapeOf({ name: v.str })) })
+    );
+
+    expect(() => validator({})).toThrowError(
+      'Validation failed! person is required'
+    );
+    expect(() => validator({ person: 9 })).toThrowError(
+      'Validation failed! person must be object'
```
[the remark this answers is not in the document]
davidtheclark: We could either write a function that adds those `return (/^[aeiou]/.test(str)) ? 'an' : 'a'`, or add it to each thing's description. (The disadvantage of the function is that it won't be 100% linguistically accurate.)

services/service-helpers/__tests__/validator3.test.js
```diff
@@ -0,0 +1,246 @@
...
+    expect(() => v.warn(v.req(v.arrayOf(v.str)))(null)).toThrowError(
+      'Validation failed! value is required'
+    );
+
+    expect(() => v.warn(v.req(v.arrayOf(v.str)))([null])).not.toThrowError();
+
+    expect(() =>
+      v.warn(v.req(v.arrayOf(v.req(v.str))))(['good', null])
+    ).toThrowError(
+      'Validation failed! item at position 1 cannot be undefined/null'
```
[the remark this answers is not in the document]
kepta: I thought required only makes sense with an object. Do you think it makes sense to say the same for an item.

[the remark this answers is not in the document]
davidtheclark: Oh, I see. Ok 👍

services/service-helpers/__tests__/validator3.test.js
```diff
@@ -0,0 +1,246 @@
+'use strict';
+
+const v = require('../validator3').v;
+
+describe('v.shapeOf', () => {
+  test('plain object', () => {
+    expect(() => v.warn(v.shapeOf({ name: v.req(v.str) }))({})).toThrowError();
```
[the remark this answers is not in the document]
kepta: ah! thanks for flagging that. This would definitely make tests easier to write and assert.

services/service-helpers/__tests__/validator3.test.js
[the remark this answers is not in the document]
kepta: neat!

services/service-helpers/validator3.js
kepta: I added some details in the `v.warn` function https://github.com/mapbox/mapbox-sdk-js/pull/230/commits/197ea1ebeb68e31b98827ebc0f028d3244037182#diff-70b24b5c23111370e645a4f4e0fb962eR24

kepta:
@davidtheclark once you review this PR and https://github.com/mapbox/mapbox-sdk-js/pull/232, merging #232  onto this should fix the failing test for this PR.

Side Note: Feel free to fix typos/bugs if any and merge it. I really don't want this PR to continue blocking other work.

davidtheclark:
@kepta Brilliant work. I have just a few small suggestions, which are a little finicky but affect the interface so might be worth addressing up front:

- `shape` is more familiar (prop-types, yup) than `shapeOf`, and I think it might fit the usage a little more precisely. The value must be a `shape`, kind of like it must be a `string` or `boolean`.
- Can we just export one object? `const v = require('./validator').v` is a little annoying, and `validate` could be attached to the same object, right? — accessible as `v.validate`?
- Since we will *always* repeat `v.warn(v.shapeOf(..))` in this library, it would be nice to have a shortcut. Do you think it would make sense to have a `validateShape()` function (whether you think that's appropriate for the library or if it should be a wrapper utility)?
- Instead of `validate` and `warn`, what do you think of `validate` (returns an array of messages) and `assert` (throws an error). My suggested `validateShape()` above could be changed to `assertShape()`. 
- How much harder do you think it would be to accumulate *all* messages and group them together instead of throwing an error with just the first one? (If an error is thrown, it could have a `messages` or `validationErrors` array or something attached to it.) Not critical if you think it would be a big pain; but if it wouldn't be too bad, I think it would be nice for a user to see and fix all the errors instead of moving through them one at a time.

Please consider these suggestions but then make the decisions based on your own best judgment: you're in charge — I'm just throwing out ideas. When you feel it's ready, feel free to merge your branches and move forward.

I will be distracted on other things until next Tuesday, so don't worry about blocking me.

kepta:
> How much harder do you think it would be to accumulate all messages and group them together 

image

@davidtheclark this is how I displaying multiple errors in an object. Let me know if there is a better way to word/format it. 

> Since we will always repeat v.warn(v.shapeOf(..)) in this library,

I think this can be a customisation done by sdk-js.In my recent commits I have added this helper function `assertShape`. I have also moved the validation code out into a separate folder, this will help us when we have time to move it out to a separate package.

davidtheclark:
@kepta 👍 Nice work. If you feel confident that the API won't need to change significantly (error messages can easily be revisited), let's merge this and finish up the services.

kepta:
Will go ahead and merge. We can revisit fine tuning error messages later.

PR #231, opened 2018-05-23
Geocoding service

davidtheclark:
The basics are in place, but this PR is dependent on @kepta's pending changes to the validator.

### DIFF
--- a/services/geocoding.js
+++ b/services/geocoding.js
@@ -1,0 +1,107 @@
+'use strict';
+
+var v = require('./service-helpers/validator');
+var pick = require('./service-helpers/pick');
+var createServiceFactory = require('./service-helpers/create-service-factory');
+
+/**
+ * Geocoding API service.
+ */
+var Geocoding = {};
+
+/**
+ * Search for a place.
+ *
+ * See the [public documentation](https://www.mapbox.com/api-documentation/#search-for-places).
+ *
+ * @param {Object} config
+ * @param {string} config.query - A place name.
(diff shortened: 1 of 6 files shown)


kepta:
just updated the validator part.

davidtheclark:
@kepta This is now ready for review.

services/geocoding.js
```diff
@@ -0,0 +1,101 @@
...
+ * @param {string|Array<string>} [config.language]
+ * @return {MapiRequest}
+ */
+Geocoding.forwardGeocode = function(config) {
+  v.assertShape({
+    query: v.required(v.string),
+    mode: v.required(v.oneOf('mapbox.places', 'mapbox.places-permanent')),
+    country: v.oneOfType(v.string, v.arrayOf(v.string)),
+    proximity: v.coordinates,
+    types: v.arrayOf(v.string),
```
kepta: We could check for all options here `Options are  country ,  region ,  postcode ,  district , place ,  locality ,  neighborhood ,  address ,  poi , and  poi.landmark .`

davidtheclark:
@kepta I added `oneOf` validation for the `type` options. In order to avoid repeating all those strings I used an array as the argument to `oneOf()`, which meant a little change to `validator.js`. Given this use case that makes arrays useful, I wonder if we should consider expecting arrays always instead of allow either argument lists *or* arrays. I know you're often interested in cutting down API surface 😄 What do you think?

davidtheclark:
I'm going to merge this PR, and we can consider changes to the validator as follow-up tasks.

PR #234, opened 2018-05-29
Map matching

kepta:
- Adds map matching
- Adds matrix

### DIFF
--- a/services/matching.js
+++ b/services/matching.js
@@ -97,7 +97,7 @@ Matching.getMatching = function(config) {
       // avoid sending params which are all `;`
       if (
         matchPath[prop].every(function(value) {
-          return value == '';
+          return value === '';
         })
       ) {
         delete matchPath[prop];
@@ -113,7 +113,7 @@ Matching.getMatching = function(config) {
 
   if (
     matchPath.isWaypoint.every(function(value) {
-      return value == true;
+      return value === true;
     })
   ) {
     delete matchPath.isWaypoint;
@@ -121,7 +121,7 @@ Matching.getMatching = function(config) {
     // the api requires the indexes to be sent
     matchPath.isWaypoint = matchPath.isWaypoint
       .map(function(val, i) {
-        return val == true ? i : '';
+        return val === true ? i : '';
       })
       .join(';');
   }
--- a/services/matching.js
+++ b/services/matching.js
@@ -3,6 +3,8 @@
 var v = require('./service-helpers/validator');
 var createServiceFactory = require('./service-helpers/create-service-factory');
 var pick = require('./service-helpers/pick');
+var objectMap = require('./service-helpers/object-map');
+var urlUtils = require('../lib/helpers/url-utils');
 
 /**
  * Map Matching API service.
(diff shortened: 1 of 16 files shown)


services/matching.js
```diff
@@ -0,0 +1,156 @@
...
+      }
+    });
+  });
+
+  ['coordinates', 'approach', 'radius', 'waypointName', 'timestamp'].forEach(
+    function(prop) {
+      // avoid sending params which are all `;`
+      if (
+        matchPath[prop].every(function(value) {
+          return value == '';
```
davidtheclark: Can we stick with the more explicit `===` by default, except maybe for the case of `x == null` (null/undefined)? I find myself wondering whether the intention with the `==` is to perform coercion or not.

services/matching.js
```diff
@@ -0,0 +1,156 @@
...
+    tidy: config.tidy,
+    approaches: matchPath.approach,
+    radiuses: matchPath.radius,
+    waypoints: matchPath.isWaypoint,
+    timestamps: matchPath.timestamp,
+    waypoint_names: matchPath.waypointName
+  };
+
+  return this.client.createRequest({
+    method: 'GET',
```
davidtheclark: Do you think we might want to use the `POST` method, in order to avoid problems with URLs getting too long? https://www.mapbox.com/api-documentation/#using-http-post

kepta: I tried using the `post` method, but it seems to not work. The server doesn't seem to understand `form-data` post requests. 
When I try to add the header `"content-type", "application/x-www-form-urlencoded"`, the header shows up in the chrome network inspection tab, but the post request is still made with 
image

Any idea what I might be missing?

kepta: Here is the code snippet I used
```javascript
tryServiceMethod('matching', 'getMatching', {"matchPath": [{"coordinates":[-117.1728265285492,32.71204416018209]},{"coordinates":[-117.17288821935652,32.712258556224]},{"coordinates":[-117.17293113470076,32.712443613445814]},{"coordinates":[-117.17292040586472,32.71256999376694]},{"coordinates":[-117.17298477888109,32.712603845608285]},{"coordinates":[-117.17314302921294,32.71259933203019]},{"coordinates":[-117.17334151268004,32.71254065549407]}]
},accessToken , console.log)
```

kepta: If I am getting it correctly, we need to implement `form-urlencoded` option in `MapRequest`.

kepta:
@davidtheclark this is ready for review.

services/matching.js
```diff
@@ -0,0 +1,171 @@
...
+      approaches: matchPath.approach,
+      radiuses: matchPath.radius,
+      waypoints: matchPath.isWaypoint,
+      timestamps: matchPath.timestamp,
+      waypoint_names: matchPath.waypointName,
+      coordinates: matchPath.coordinates
+    },
+    function(_, val) {
+      return val != null;
+    }
```
davidtheclark: Seems like we're be using this pattern enough that it would be worth having a dedicated shortcut function, `cleanObject`? That way we can be sure we have the same implementation in each place (e.g. removing *both* `undefined` and `null` but not other falsy values).

services/matching.js
```diff
@@ -0,0 +1,171 @@
...
+    geometries: v.oneOf('geojson', 'polyline', 'polyline6'),
+    language: v.string,
+    overview: v.oneOf('full', 'simplified', 'false'),
+    steps: v.boolean,
+    tidy: v.boolean
+  })(config);
+
+  config.profile = config.profile || 'driving';
+
+  var matchPath = {
```
davidtheclark: Can we give this a name that distinguishes it from `config.matchPath`?

services/matching.js
```diff
@@ -0,0 +1,171 @@
...
+      timestamps: matchPath.timestamp,
+      waypoint_names: matchPath.waypointName,
+      coordinates: matchPath.coordinates
+    },
+    function(_, val) {
+      return val != null;
+    }
+  );
+  body = objectMap(body, function(_, value) {
+    // appendQueryObject omits boolean values
```
davidtheclark: Does really "omit" them — just doesn't stringify the boolean value.

services/matrix.js
```diff
@@ -0,0 +1,101 @@
...
+    ),
+    profile: v.oneOf('driving-traffic', 'driving', 'walking', 'cycling'),
+    annotations: v.arrayOf(v.oneOf('duration', 'distance')),
+    sources: v.oneOfType(v.equal('all'), v.arrayOf(v.number)),
+    destinations: v.oneOfType(v.equal('all'), v.arrayOf(v.number))
+  })(config);
+
+  config.profile = config.profile || 'driving';
+
+  var matrixPath = {
```
davidtheclark: Same here: let's distinguish this from the user input

services/matrix.js
```diff
@@ -0,0 +1,101 @@
...
+        })
+      )
+    ),
+    profile: v.oneOf('driving-traffic', 'driving', 'walking', 'cycling'),
+    annotations: v.arrayOf(v.oneOf('duration', 'distance')),
+    sources: v.oneOfType(v.equal('all'), v.arrayOf(v.number)),
+    destinations: v.oneOfType(v.equal('all'), v.arrayOf(v.number))
+  })(config);
+
+  config.profile = config.profile || 'driving';
```
davidtheclark: The API requires this value instead of providing a default. Do you think we should provide *additional* defaults like this in the SDK, or require all the input that the HTTP API requires?

I ask because I was thinking about this briefly for Geocoding: it would make sense to default the `mode` to `mapbox.places` (instead of `mapbox.places-permanent`), but the API didn't do that so I was hesitant.

What do you think?

kepta: I feel the directions profile is special option, since it goes in the url. I can make it required 🤷🏽‍♀️

davidtheclark: Whatever you want to go with is ok with me. Let's at some point make sure that Geocoding matches the pattern.

kepta: Went ahead and [updated](https://github.com/mapbox/mapbox-sdk-js/pull/234/commits/82f4cff2440e7c4bdf7c18535b761ed1e23f6fb7) geocoding service to use a default mode.

PR #235, opened 2018-05-31
add upload service

kepta:
Adds uploads service.

### DIFF
--- a/services/uploads.js
+++ b/services/uploads.js
@@ -1,0 +1,116 @@
+'use strict';
+
+var v = require('./service-helpers/validator');
+var createServiceFactory = require('./service-helpers/create-service-factory');
+
+/**
+ * Uploads API service.
+ */
+var Uploads = {};
+
+/**
+ * List all recent upload statuses
+ *
+ * See the [public documentation](https://www.mapbox.com/api-documentation/#uploads).
+ *
+ * @param {Object} [config]
+ * @param {boolean} [config.reverse] - List uploads in chronological order, rather than reverse chronological order.
+ * @return {MapiRequest}
+ */
+Uploads.listUploads = function(config) {
+  v.assertShape({
+    reverse: v.boolean
+  })(config);
+
+  return this.client.createRequest({
+    method: 'GET',
+    path: '/uploads/v1/:ownerId',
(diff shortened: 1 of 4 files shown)


services/uploads.js
```diff
@@ -0,0 +1,116 @@
...
+  });
+};
+
+/**
+ * Create an upload.
+ *
+ * See the [public documentation](https://www.mapbox.com/api-documentation/#create-an-upload).
+ *
+ * @param {Object} [config
+ * @param {string} config.tileset - the map ID to create or replace in the format  username.nameoftileset - limited to 32 characters (only  - and  _ special characters allowed, limit does not include username)
```
davidtheclark: Let's call this `mapId` — I think that pretty consistently should refer to the `{ownerId}.{tilesetId}` pairing, whereas `tileset` is kind of ambiguous.

services/uploads.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {string} config.url - HTTPS URL of the S3 object provided in the credential request or the dataset ID of an existing Mapbox dataset to be uploaded
```
davidtheclark: How about `s3Url`, to make it clear that this is not the URL of the request you're making?

services/uploads.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {string} [config.name] - name for the tileset - limited to 64 characters
```
davidtheclark: How about `tilesetName`, so it's clear you're naming the tileset not the upload?

services/uploads.js
```diff
@@ -0,0 +1,116 @@
...
+    path: '/uploads/v1/:ownerId/credentials'
+  });
+};
+
+/**
+ * Create an upload.
+ *
+ * See the [public documentation](https://www.mapbox.com/api-documentation/#create-an-upload).
+ *
+ * @param {Object} [config
```
davidtheclark: Looks like there stray `[`s here and below.

kepta:
Yup I tried it out, but I used the dataset instead of s3. Which I feel should be the same ?

davidtheclark:
👍

PR #236, opened 2018-06-04
Static API

davidtheclark:
The Static API is very unique, tricky, weird ...

@kepta I hope to finish this tomorrow.


davidtheclark:
@kepta I guess this is ready for review after all, because we don't need to add the other Static API service methods right (they are not in the legacy SDK).

services/static.js
```diff
@@ -0,0 +1,245 @@
...
+ * @return {MapiRequest}
+ */
+Static.getStaticImage = function(config) {
+  v.assertShape({
+    ownerId: v.required(v.string),
+    styleId: v.required(v.string),
+    width: v.required(v.number),
+    height: v.required(v.number),
+    coordinates: v.required(v.oneOfType(v.coordinates, v.oneOf('auto'))),
+    zoom: v.required(v.number),
```
kepta: we can use range for a bunch of props here

davidtheclark: Good idea 👍

PR #237, opened 2018-06-05
Normalize some things and improve documentation

davidtheclark:
I'll add to this list of changes when I finish tomorrow:

- `Datasets#getDataset > Datasets#getMetadata` and `Datasets#updateDataset > Datasets#updateMetadata`, because you're really dealing with the metadata in those methods.
- `directionsPath > waypoints` and `DirectionsPath > DirectionsWaypoint`
- `Matching > MapMatching` , to match product name.
- `matchPath > points` and `MatchPath > MapMatchingPoint`
- `MapMatching` item `timestamp` can be any valid date, not just a Unix timestamp.
- Added `services/service-helpers/generic-types.js`
- Arrays required for Geocoding's `country` and `language` options.
- `MatrixPath > MatrixPoint` and `matrixPath > points`
- In Tilequery, `mapIds` must always be an array (used to be `mapId`).
- In Tilequery, `config.longitude` and `config.latitude` are condensed into `config.coordinates`.
- In Static, `overlays` must always be an array (used to be `overlay`).
- In Tokens, removed `ownerId` config option, since I don't think that can be used.


davidtheclark:
@kepta ready for review.

Issue #239, opened 2018-06-07
New validation lib doesn't reject unwanted properties


davidtheclark:
@kepta I noticed that the new validation lib does not reject unwanted properties. For example, I can send a request to `Datasets#listDatasets` with a `waypoints` property, and there's no error.

Do you think we should make it more strict, so *only* properties specified in the schema are allowed? Or do you think we should keep it loose?

kepta:
Ya, I believe that was intentional on my side to follow the lose nature of `prop-type`. But I can see why that would be a problem. 
We can do something similar to airbnb's [forbidExtraProps](https://github.com/airbnb/prop-types). 

This would be another validator similar to `v.shapeOf` but which forbids any extra properties.

davidtheclark:
> that was intentional on my side

Good choice 👍 That enables you to pass around objects without validating every single property every time. I think that's a good default.

For our case in this SDK, a `strictShapeOf` would be great, so users are aware when they are passing unnecessary properties.

davidtheclark:
I added a `v.strictShape` higher-order validator to fusspot, so this issue is actionable.

PR #241, opened 2018-06-08
Tidy docs by removing empty configs and ordering UploadableFile

davidtheclark:
Because we won't currently throw validation errors if unnecessary props are provided (https://github.com/mapbox/mapbox-sdk-js/issues/239), we can just remove the validation step from these methods.

@kepta for review.


davidtheclark:
I also removed `ownerId` params from the Tokens service, since that was misleading — I'm pretty sure you can't request anybody else's tokens.

Issue #135, opened 2017-03-13
Add Optimization API


tristen:
https://www.mapbox.com/api-documentation/#optimized-trips

davidtheclark:
@tristen: Looks like Distance API was replaced with the Matrix API: https://github.com/mapbox/mapbox-sdk-js/blob/master/CHANGELOG.md#100-beta8. Should we change this issue title so it's about *adding* the Optimization API?

danswick:
Closed by #257.

PR #257, opened 2018-07-10
Add Optimization API

danswick:
Addresses https://github.com/mapbox/mapbox-sdk-js/issues/135. 

Adds Optimization API, mostly modeled after @kepta's work on the Directions API. The [`distributions`](https://www.mapbox.com/api-documentation/#retrieve-an-optimization) parameter threw me for a bit of a loop (unlike, say, `radius` or `bearing`, you don't need one `distribution` per coordinate) and I'm feeling a bit unsure about the way I'm currently handling it. 

Any and all feedback welcomed!

### DIFF
--- a/services/__tests__/optimization.test.js
+++ b/services/__tests__/optimization.test.js
@@ -1,0 +1,229 @@
+'use strict';
+
+const optimizationService = require('../optimization');
+const tu = require('../../test/test-utils');
+
+let optimization;
+beforeEach(() => {
+  optimization = optimizationService(tu.mockClient());
+});
+
+describe('getOptimization', () => {
+  test('works', () => {
+    optimization.getOptimization({
+      waypoints: [
+        {
+          coordinates: [2.2, 1.1]
+        },
+        {
+          coordinates: [2.2, 1.1]
+        }
+      ]
+    });
+    expect(tu.requestConfig(optimization)).toEqual({
+      path: '/optimized-trips/v1/mapbox/:profile/:coordinates',
+      method: 'GET',
+      params: {
+        coordinates: '2.2,1.1;2.2,1.1',
+        profile: 'driving'
+      },
+      query: {}
+    });
+  });
+
+  test('it omits queries not supplied', () => {
+    optimization.getOptimization({
+      waypoints: [
+        {
+          coordinates: [2.2, 1.1]
+        },
+        {
+          coordinates: [2.2, 1.1]
+        }
+      ],
+      profile: 'walking',
+      geometries: 'polyline'
+    });
+    expect(tu.requestConfig(optimization)).toEqual({
+      path: '/optimized-trips/v1/mapbox/:profile/:coordinates',
+      method: 'GET',
+      params: {
+        coordinates: '2.2,1.1;2.2,1.1',
+        profile: 'walking'
+      },
+      query: {
+        geometries: 'polyline'
+      }
+    });
+  });
+
+  test('it reads waypoints props', () => {
+    optimization.getOptimization({
+      waypoints: [
+        {
+          coordinates: [2.2, 1.1],
+          radius: 2000,
+          bearing: [45, 20]
+        },
+        {
+          coordinates: [2.2, 1.1],
+          radius: 2000,
+          bearing: [46, 21]
+        }
+      ],
+      profile: 'walking',
+      steps: false
+    });
+    expect(tu.requestConfig(optimization)).toEqual({
+      path: '/optimized-trips/v1/mapbox/:profile/:coordinates',
+      method: 'GET',
+      params: {
+        coordinates: '2.2,1.1;2.2,1.1',
+        profile: 'walking'
+      },
+      query: {
+        steps: 'false',
+        radiuses: '2000;2000',
+        bearings: '45,20;46,21'
+      }
+    });
+  });
+
+  test('it works if an optional waypoints.bearing is missing at some places', () => {
+    optimization.getOptimization({
+      waypoints: [
+        {
+          coordinates: [2.2, 1.1]
+        },
+        {
+          coordinates: [2.2, 1.1]
+        },
+        {
+          coordinates: [2.2, 1.1],
+          bearing: [45, 32]
+        },
+        {
+          coordinates: [2.2, 1.1]
+        }
+      ],
+      profile: 'walking',
+      steps: false
+    });
+    expect(tu.requestConfig(optimization)).toEqual({
+      path: '/optimized-trips/v1/mapbox/:profile/:coordinates',
+      method: 'GET',
+      params: {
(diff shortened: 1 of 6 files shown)


services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+ * Get a duration-optimized route.
+ *
+ * Please read [the full HTTP service documentation](https://www.mapbox.com/api-documentation/#optimization)
+ * to understand all of the available options.
+ *
+ * @param {Object} config
+ * @param {'driving'|'walking'|'cycling'} [config.profile="driving"]
+ * @param {Array<OptimizationWaypoint>} config.waypoints - An ordered array of [`OptimizationWaypoint`](#optimizationwaypoint) objects, between 2 and 12 (inclusive).
+ * @param {Array<'duration'|'distance'|'speed'>} [config.annotations] - Specify additional metadata that should be returned.
+ * @param {'any' | 'last'} [config.destination="any"] Returned route ends at `any` or `last` coordinate.
```
davidtheclark: Nitpicky formatting points — spaces around `|` and lack of `-` between name and description.

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {[number, number]} [config.distributions] Array of objects, each of which includes a `pickup` and `dropoff` property. `pickup` and `dropoff` properties correspond to an index in the coordinates array.
```
davidtheclark: Type definition looks wrong here.

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {'geojson'|'polyline'|'polyline6'} [config.geometries="polyline"] - Format of the returned geometry.
```
davidtheclark: I think it's format of the returned geometries (plural).

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+        v.shape({
+          coordinates: v.required(v.coordinates),
+          approach: v.oneOf('unrestricted', 'curb'),
+          bearing: v.arrayOf(v.range([0, 360])),
+          radius: v.oneOfType(v.number, v.equal('unlimited'))
+        })
+      )
+    ),
+    annotations: v.arrayOf(v.oneOf('duration', 'distance', 'speed')),
+    geometries: v.string,
```
davidtheclark: Looks like we could be more specific here.

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+    destination: v.oneOf('any', 'last'),
+    distributions: v.arrayOf(
+      v.shape({
+        pickup: v.number,
+        dropoff: v.number
+      })
+    )
+  })(config);
+
+  config.profile = config.profile || 'driving';
```
davidtheclark: If we've been using this pattern in other places (which we probably have), we should change it. This will mutate the object that the user passed in — which we don't want to do.

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+    coordinates: [],
+    approach: [],
+    bearing: [],
+    radius: [],
+    distributions: []
+  };
+
+  var waypointCount = config.waypoints.length;
+  if (waypointCount < 2 || waypointCount > 12) {
+    // updated to reflect opt restrictions
```
davidtheclark: I'm not sure what this comment means?

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+    } else {
+      path[prop] = path[prop].join(';');
+    }
+  });
+
+  if (config.distributions) {
+    var distributionsList = config.distributions.map(function(dist) {
+      return [dist.pickup, dist.dropoff];
+    });
+    path.distributions.push(distributionsList.join(';'));
```
davidtheclark: It's odd to me that `path.distributions` is still an array but will only ever end up with one item in it, right?

services/optimization.js
```diff
@@ -0,0 +1,162 @@
...
+    geometries: config.geometries,
+    language: config.language,
+    overview: config.overview,
+    roundtrip: config.roundtrip,
+    steps: config.steps,
+    source: config.source,
+    destination: config.destination,
+    distributions: path.distributions.length
+      ? path.distributions[0]
+      : config.distributions,
```
davidtheclark: Why would we fall back to `config.distributions`?

services/__tests__/optimization.test.js
```diff
@@ -0,0 +1,229 @@
...
+      method: 'GET',
+      params: {
+        coordinates: '2.2,1.1;2.2,1.1',
+        profile: 'driving'
+      },
+      query: {}
+    });
+  });
+
+  test('it omits queries not supplied', () => {
```
davidtheclark: Not sure what this description means ... what's being "omitted" in this test?

services/__tests__/optimization.test.js
```diff
@@ -0,0 +1,229 @@
...
+      },
+      query: {
+        steps: 'false',
+        bearings: ';;45,32;',
+        radiuses: '2000;;;'
+      }
+    });
+  });
+
+  test('waypoints.radius can be any of string or number', () => {
```
davidtheclark: Not really any string, right?

services/__tests__/optimization.test.js
```diff
@@ -0,0 +1,229 @@
...
+      method: 'GET',
+      params: {
+        coordinates: '2.2,1.1;2.2,1.1;2.2,1.1;2.2,1.1',
+        profile: 'driving'
+      },
+      query: {
+        distributions: '0,1;2,3'
+      }
+    });
+  });
```
davidtheclark: Can we test the error condition if you provide the wrong number of waypoints?

services/optimization.js
danswick: Moved this check down to `createRequest` instead:

```
params: {
      profile: config.profile || 'driving',
      coordinates: path.coordinates.join(';')
    }
```

services/optimization.js
danswick: Oops! That was a note I left for myself. Removed!

services/optimization.js
danswick: Thanks for this! Per discussion: 

1. Moved the `join` piece to the `forEach` loop along with `approach`, `bearing`, etc. 
2. Pushed each pickup/dropoff pair to `path.distributions` instead of putting them into various unnecessary arrays first.

services/optimization.js
danswick: Checking for `config.distributions` while doing the `.join(';')` dance above makes this unnecessary, so I removed it!

services/__tests__/optimization.test.js
danswick: Updated the wording here, but I'm thinking we might not even need this test since most of the others will fail if a stray query gets added anyway.

davidtheclark:
Let's remove the inadvertent `package-lock.json` change, then we can merge and cut a new release.

PR #264, opened 2018-07-25
Update Fusspot, use local date validator

davidtheclark:
Follow-up to https://github.com/mapbox/fusspot/issues/5

@kepta for review, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## HEAD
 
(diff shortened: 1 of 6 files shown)


davidtheclark:
We'll want to wait on a release of this until https://github.com/mapbox/fusspot/pull/15 is merged and released in Fusspot.

davidtheclark:
Added another commit that introduces a local `v.coordinates` validator (since the generic one was removed from Fusspot), which only enforces that you provide two numbers. It lets the API validate your numbers.

PR #267, opened 2018-07-30
Reverse coordinates before sending them to polyline.encode

### LINKED ISSUE #266: mapboxStatic path broken for lng < -90
Hello,

When I try to add a path that extends beyond -90 lng, the client errors:

Here is request I'm attempting that errors:

```
.getStaticImage({
        ownerId: 'mapbox',
        styleId: 'streets-v10',
        width: 300,
        height: 300,
        coordinates: [-80, 41],  // [ lng, lat ]
        zoom: 2,
        overlays: [
          {
            path: {
              /*
                - these coordinates need to be reversed to lat / lng for the path to draw
                - this is the opposite of the other coordinates ordering
                - relates to https://github.com/mapbox/mapbox-sdk-js/issues/256
              */
              coordinates: [
                [41, -70],  // [ lat,  lng ]
                [41, -90.0001], // [ lat,  lng ]:  errors for all values < -90 (-90.001, -100, -107, etc.)
              ],
            },
          },
        ],
      });
```

and the error:

```
client err Error: Item at position coordinates.1 must be an array of [longitude, latitude].
```

The following request works and results in the attached map:

```
.getStaticImage({
        ownerId: 'mapbox',
        styleId: 'streets-v10',
        width: 300,
        height: 300,
        coordinates: [-80, 41],  // [ lng, lat ]
        zoom: 2,
        overlays: [
          {
            path: {
              /*
                - these coordinates need to be reversed to [ lat,  lng ] for the path to draw
                - this is the opposite of the other coordinates ordering
                - relates to https://github.com/mapbox/mapbox-sdk-js/issues/256
              */
              coordinates: [
                [41, -70],  // [ lat,  lng ]
                [41, -90],  // [ lat,  lng ]
              ],
            },
          },
        ],
      });;
```

image

reversing the coordinates[1] per the error looks like this: 

```
      .getStaticImage({
        ownerId: 'mapbox',
        styleId: 'streets-v10',
        width: 300,
        height: 300,
        coordinates: [-80, 41],  // [ lng, lat ]
        zoom: 2,
        overlays: [
          {
            path: {
              /*
                - these coordinates need to be reversed to lat / lng for the path to draw
                - this is the opposite of the other coordinates ordering
                - relates to https://github.com/mapbox/mapbox-sdk-js/issues/256
              */
              coordinates: [
                [41, -70],  // [ lat,  lng ]
                [-90.001, 41],  // [ lng,  lat ]
              ],
            },
          },
        ],
      });
```

image

This works in that I get the line, but the coordinates[1] indexes are out of order so the line doesn't terminate where I want.

davidtheclark:
Closes #266.

@danswick for review, please. The details in #266 are helpful for manually reproducing the original bug and verifying the fix.

### DIFF
--- a/services/static.js
+++ b/services/static.js
@@ -221,7 +221,11 @@ function encodePathOverlay(o) {
   if (o.fillOpacity) {
     result += '-' + o.fillOpacity;
   }
-  var encodedPolyline = polyline.encode(o.coordinates);
+  // polyline expects each coordinate to be in reversed order: [lat, lng]
+  var reversedCoordinates = o.coordinates.map(function(c) {
+    return c.reverse();
+  });
+  var encodedPolyline = polyline.encode(reversedCoordinates);
   result += '(' + encodeURIComponent(encodedPolyline) + ')';
   return result;
 }
--- a/services/static.js
+++ b/services/static.js
@@ -176,7 +176,7 @@ function encodeCustomMarkerOverlay(o) {
  * A stylable line.
  * @typedef {Object} PathOverlay
  * @property {Object} path
- * @property {Array<[number, number]>} path.coordinates - An array of coordinates
+ * @property {Array<Coordinates>} path.coordinates - An array of coordinates
  *   describing the path.
  * @property {number} [path.strokeWidth]
  * @property {string} [path.strokeColor]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## HEAD
+
+- **Fix:** Fix bug in `Static#getStaticImage` that resulted in reversed coordinates when creating a polyline overlay.
+
 ## 0.3.0
 
 - **Change:** Rename `MapMatching#getMatching` to `MapMatching#getMatch`.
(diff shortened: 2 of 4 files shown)


chrbradley:
@davidtheclark 

Thanks a ton for the quick turnaround!

This will work with lng < - 90 because checks on the coordinates will now see them in the right order?

davidtheclark:
@chrbradley Yep, it should work. Your arguments to the function should fit the typical (and currently documented) `[lng,lat]` convention (so a longitude of -91 should not be rejected), and behind the scenes they'll be reversed before they're passed to polyline. 

I did try your parameters with this branch and got the straight line under the Great Lakes that you were expecting.

services/static.js
```diff
@@ -176,7 +176,7 @@ function encodeCustomMarkerOverlay(o) {
  * A stylable line.
  * @typedef {Object} PathOverlay
  * @property {Object} path
- * @property {Array<[number, number]>} path.coordinates - An array of coordinates
+ * @property {Array<Coordinates>} path.coordinates - An array of coordinates
```
danswick: Added a small change here to ensure it's clear that these input coordinates should follow the same `[lng, lat]` format.

davidtheclark: Ah, thanks.

PR #270, opened 2018-08-11
Fix CORS bug when combining Firefox and the Tilequery API

davidtheclark:
We noticed a bug where Firefox was sending preflight `OPTIONS` requests to the Tilequery API that the server was rejecting. 

The Tilequery backend is not currently configured to handle CORS because it doesn't really *need* to: its work can be done via ["simple requests"](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests). We thought the SDK was sending "simple requests" when possible, so avoiding unnecessary preflight requests — in Chrome and some other browsers it *was*. But at the bottom of the list of things that make requests non-simple there's this trick:

> one or more event listeners are registered on an XMLHttpRequestUpload object used in the request.

Most browsers were ignoring this detail in the spec, whereas Firefox was not. And most of our APIs were avoiding the Firefox bug because they handled CORS preflight requests, whereas Tilequery was not. Put those together and you get an error.

This PR remedies the situation by only attaching upload event listeners if the request includes a file — in which case it's not going to be "simple" anyway.

The current bad behavior can be reproduced by using `npm run try-browser`, opening the page in Firefox, and (setting a public access token) running this:

```
tryServiceMethod('tilequery', 'listFeatures', {
  "mapIds": ["mapbox.mapbox-streets-v7"],
  "coordinates": [-122.42901,37.80633]
}).send()
```

The same request will work in Firefox now.

@kepta for review, please.

cc @mapsam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## HEAD
+
+- **Fix:** Fix a CORS-related bug that caused Firefox to send preflight `OPTIONS` requests that were rejected by the server. This bug surfaced in Firefox with the Tilequery API, but may possibly have affected some other endpoints and some other browsers.
+
 ## 0.4.0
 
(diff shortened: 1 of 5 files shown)


andrewharvey:
Same issue on IE11, confirmed it's fixed with this :+1: .

davidtheclark:
Thanks for the confirmation, @andrewharvey!

PR #272, opened 2018-08-15
hasNextPage readme example

mapsam:
This updates the `eachPage()` docs to include the `res.hasNextPage()` method as an example of how to check for a next page before calling `next()`, otherwise there's no current way to determine when all pages have been returned.

cc @danswick

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,11 @@ req.send().then(response => {..}, error => {
 req.abort();
 
 // Paginate through a response.
-tilesetsService.listTilesets().eachPage((error, response, next) => {
+tilesetsService.listTilesets().eachPage((error, response, next) => {  
   // Do something with the page, then call next() to send the request
-  // for the next page.
+  // for the next page. If you want to check if there are any more pages to call
+  // use res.hasNextPage() first.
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks, @mapsam. Do you have any opinions about whether it would be preferable to have `eachPage` call your callback with `null` when you have exhausted the pages, or to instead document this pattern?

(I notice that the [AWS JS SDK's `eachPage`](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#eachPage-property) does the former, calling the callback with `null`.)

andrewharvey:
+1 I'd been checking `response.links.next`, so having the proper way to do this in the example would be good.

mapsam:
@davidtheclark nope, no opinion from me. Thanks for asking though! I'll lean on your preference - and be happy with it being documented.

davidtheclark:
Thanks @mapsam and @andrewharvey for the input. I'll merge this now and open a new issue for consideration some API changes that could make this use case a little smoother.

mapsam:
Thanks @davidtheclark & @danswick!

PR #277, opened 2018-08-18
include origin option in MapiClient

andrewharvey:
`MapiClient` `options.origin` is part of the public API, this PR ensures it's documented. One use case for this is changing it to `api.mapbox.cn`.


davidtheclark:
Thanks, @andrewharvey.

Issue #289, opened 2018-11-16
Verify Map Matching service


davidtheclark:
[A recent issue](https://github.com/mapbox/mapbox-sdk-js/issues/286) about Map Matching spurred me to look at the HTTP API docs, and I found some irregularities. Those have been fixed, so I think we should check whether those fixes affect this library's Map Matching service — whether any corresponding fixes need to be made here.

@danswick is this something you'd like to take on?

danswick:
Looks like the example code in the HTTP API docs is not quite right. I get `Waypoints must be integer indexes that reference a coordinate` in response from the map-matching API (weirdly, I'm unable to find a reference to this error code anywhere in the directions API codebase). 

I noticed a couple of issues that I can take a crack at addressing this week: 

1. It looks like we're constructing requests with multiple semicolons when there are instances of `isWaypoint: false` in the `points` array. 
2. [api docs issue] we state that the number of timestamps must be the same as the number of coordinates, but the example code just has one timestamp.

PR #295, opened 2018-12-19
[WIP] Update map matching to filter empty waypoints

danswick:
Addresses part of https://github.com/mapbox/mapbox-sdk-js/issues/289. 

While verifying the map-matching service, I noticed we weren't filtering empty waypoints, which would result in a `'Waypoints must be integer indexes that reference a coordinate'` error message from the Map Matching API. This PR adds a filter to make sure we don't get any pesky stray `;` in our `waypoints` list. 

@kepta for review, please 🙏 

todo: 

- [x] Update tests

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -247,7 +247,7 @@ SDK returned.
   - `config.position` **(`"auto"` \| [Object][112])** If `"auto"`, the viewport will fit the
       bounds of the overlay(s). Otherwise, the maps' position is described by an object
       with the following properties:
-      `coordinates` (required): `[longitude, latitude]` for the center of image.
+      `coordinates` (required): [`coordinates`][106] for the center of image.
       `zoom` (required): Between 0 and 20.
       `bearing` (optional): Between 0 and 360.
       `pitch` (optional): Between 0 and 60.
--- a/docs/services.md
+++ b/docs/services.md
@@ -377,7 +377,7 @@ SDK returned.
   - `config.position` **(`"auto"` \| [Object][150])** If `"auto"`, the viewport will fit the
       bounds of the overlay(s). Otherwise, the maps' position is described by an object
       with the following properties:
-      `coordinates` (required): [`coordinates`][106] for the center of image.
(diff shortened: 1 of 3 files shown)


danswick:
Looks like we may also need to verify similar behavior for the following, though I'm not sure yet if the Map Matching API will be cranky about `;;` in the same way it is for `waypoints`.  

- [ ] approaches
- [ ] radiuses
- [ ] timestampes
- [ ] waypoint names

services/map-matching.js
```diff
@@ -133,6 +133,9 @@ MapMatching.getMatch = function(config) {
       .map(function(val, i) {
         return val === true ? i : '';
       })
+      .filter(function(x) {
+        return x === 0 || Boolean(x);
```
kepta: What exactly do we want to filter out in this @danswick ?

davidtheclark:
I think the key misunderstanding here is that there's not supposed to an equal number of waypoints and coordinates — right?

danswick:
@davidtheclark @kepta I believe there _can_ be an equal number of waypoints and coordinates, but the issue arises when they are not the same and the SDK joins lonely semicolons into the list of waypoints like: `0;;2;3`, in which case the Map Matching API will throw the error mentioned in the OP. So we need to filter out any `path` items that say `isWaypoint: false` (and also avoid accidentally filtering out the `0` index).

docs/services.md
```diff
@@ -377,7 +377,7 @@ SDK returned.
   - `config.position` **(`"auto"` \| [Object][150])** If `"auto"`, the viewport will fit the
       bounds of the overlay(s). Otherwise, the maps' position is described by an object
       with the following properties:
-      `coordinates` (required): [`coordinates`][144] for the center of image.
+      `coordinates` (required): [`coordinates`][106] for the center of image.
```
colleenmcginnis: Why is this changing?

danswick: good catch! This PR got old enough that I needed to catch up to master and update the docs.

PR #303, opened 2019-01-11
Docs about creating a service

davidtheclark:
Based on some conversations I've had recently, I think this doc would be helpful. 

@danswick for review, please.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,7 +27,7 @@ The properties of a service prototype object are the service methods. Each servi
 - `path` (string): the path of the endpoint, with `:express-style` colon-prefixed parameters for path parts that should be replaced by values in the `params` object.
 - `params` (object): an object whose keys correspond to the `:express-style` colon-prefixed parameters in the `path` string, and whose values are the values that should be substituted into the path. For example, with `path: '/foo/:bar/:baz'` you'll need a params object with `bar` and `baz` properties, like `{ bar: 'a', baz: 'b' }`. **You do *not* need to specify an `ownerId` param: that is automatically provided by the `MapiClient`.**
 - `query` (object): an object that will be transformed into a query string and attached to the `path` (e.g. `{ foo: 'a', baz: 'b' }` becomes `?foo=a&baz=b`).
-- `headers` (object): an object of headers that should be added to the request. Keys should be lowercase.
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -16,3 +16,48 @@
...
+
+## Creating a service
+
+First you'll create a service prototype object, then you'll pass that prototype object into the [`createServiceFactory()`](../services/service-helpers/create-service-factory.js) function.
+
+The properties of a service prototype object are the service methods. Each service method is a function that accepts a configuration object and returns a `MapiRequest`. The request should be created using `this.client.createRequest()`, which accepts an object with the following properties:
+
+- `method` (string): the HTTP method (e.g. `GET`).
+- `path` (string): the path of the endpoint, with `:express-style` colon-prefixed parameters for path parts that should be replaced by values in the `params` object.
+- `params` (object): an object whose keys correspond to the `:express-style` colon-prefixed parameters in the `path` string, and whose values are the values that should be substituted into the path. For example, with `path: '/foo/:bar/:baz'` you'll need a params object with `bar` and `baz` properties, like `{ bar: 'a', baz: 'b' }`. **You do *not* need to specify an `ownerId` param: that is automatically provided by the `MapiClient`.**
```
danswick: this is really nicely explained!

docs/development.md
```diff
@@ -16,3 +16,48 @@
...
+- `query` (object): an object that will be transformed into a query string and attached to the `path` (e.g. `{ foo: 'a', baz: 'b' }` becomes `?foo=a&baz=b`).
+- `headers` (object): an object of headers that should be added to the request. Keys should be lowercase.
```
danswick: Are there any headers provided by default? Are there any that should be included in `headers` every time?

docs/development.md
```diff
@@ -16,3 +16,48 @@
...
+- `body` (object, default `null`): a body that should be included with the `POST` or `PUT` request. It will be stringified as JSON.
```
danswick: I always get tripped up stringified wording. Should this be "stringified JSON" (without the "as")? Or is "stringified as JSON" more standard?

kepta: @danswick every JSON is a string in Javascript. In this case it means an Object (not a string ) -> would be stringified to become a JSON.

docs/development.md
davidtheclark: I'll add a note about `content-type: application/json` being automatically included if the request includes a `body`.

PR #308, opened 2019-02-15
Allow referrers and resources to have  null values

kepta:
This allow for passing a `null` value to `resources` or `referrers` properties in `Tokens#updateToken` to delete the respective property.

@davidtheclark or @danswick for review, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## HEAD
+
+- **Fix:** `Tokens#updateToken` can now set `null` value to `referrers` property to delete the property.
+- **Fix:** `Tokens#updateToken` can now set `null` value to `resources` property to delete the property.
+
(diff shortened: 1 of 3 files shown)


davidtheclark:
Turning this over to @danswick for review.

danswick:
@kepta I am getting a 500 internal server error while testing this change, but I may be working from some very wrong assumptions. Let's chat next week and maybe you can help me make sure I'm sending a proper request.

kepta:
> @kepta I am getting a 500 internal server error while testing this change, but I may be working from some very wrong assumptions

did you get a chance to see server logs to what the 500 actually was ?

danswick:
Circling back: we tracked this down and I was just making a malformed request to the tokens endpoint. However, we did uncover an unrelated bug and were able to verify that this change is working as expected, so all is not lost.

PR #310, opened 2019-03-01
Swap all references to `referrers` to `allowedUrls`

danswick:
As part of our impending token restrictions launch, we're transitioning our terminology from "referrers" to "allowedUrls". This PR swaps all references to `referrer` or `referrers` to `allowedUrl` or `allowedUrls`. 

@kepta for review, please 🙏 

cc @lshig

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - **Fix:** `Tokens#updateToken` can now set `null` value to `referrers` property to delete the property.
 - **Fix:** `Tokens#updateToken` can now set `null` value to `resources` property to delete the property.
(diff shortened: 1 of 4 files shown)


danswick:
@kepta - do you think we should support both `referrers` **and** `allowedUrls` here as well?

kepta:
> @kepta - do you think we should support both referrers and allowedUrls here as well?

Good question. To me it is looking like more effort to prevent the breaking change from happening (expressed similar concerns in https://github.com/mapbox/core-users/pull/64#issuecomment-468843741). Let us just go ahead with `allowedUrl` and introduce a breaking change :).

danswick:
Agreed, @kepta. Anyone using a previous version can continue doing so until they're ready to update 👍

PR #318, opened 2019-05-23
Update Static API to Static Images API

colleenmcginnis:
Changes references of `Static API` to `Static Images API` and updates links to docs subdomain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # @mapbox/mapbox-sdk
 
-A JS SDK for working with [Mapbox APIs](https://www.mapbox.com/api-documentation/).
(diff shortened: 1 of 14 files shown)


danswick:
👋 Heyo, @colleenmcginnis! `services.md` is generated by the [document](https://github.com/mapbox/mapbox-sdk-js/blob/master/package.json#L23) script, so unfortunately it will get overwritten the next time someone runs it. It gets all of those URLs from the inline JSDoc comments, [like this one for static](https://github.com/mapbox/mapbox-sdk-js/blob/master/services/static.js#L12), so all of those will need to be updated 😬.

colleenmcginnis:
🆗 I fixed it @danswick.

PR #327, opened 2019-08-09
Add routing as an option on geocoding

RainierXS:
noticed the API has a routing option on geocoding that was not in the SDK, preventing responses that contain the routable_points


andrewharvey:
Hi @RainierXS looks like you've done a merge rather than rebase and force push. We'll need to have just your commits rebased on top of master to merge this. If you are having trouble with the rebase then you could try taking the changed files and starting fresh on a new branch from master and force pushing that here.

PR #330, opened 2019-10-23
Add Static Images API style params: addlayer, setfilter, layer_id

katydecorah:
This PR adds the new style parameters to the Static Images API service:
* addlayer
* setfilter
* layer_id

🚨This PR also contains a breaking change: `insertOverlayBeforeLayer` to `before_layer` to make the wording consistent with the documentation, especially since addlayer can use `before_layer`.

~~This PR also makes objects passable to MapiRequest#url.~~

Tagging @riastrad for additional review.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -491,7 +491,10 @@ const request = staticClient
     styleId: 'streets-v11',
     width: 200,
     height: 300,
-    position: 'auto',
+    position: {
+      coordinates: [12, 13],
+      zoom: 4
+    },
     setfilter: [">","height",300],
     layer_id: 'building',
   });
@@ -507,9 +510,12 @@ const request = staticClient
     styleId: 'streets-v11',
     width: 200,
     height: 300,
-    position: 'auto',
+    position: {
+      coordinates: [12, 13],
+      zoom: 4
+    },
     addlayer: {"id":"better-boundary","type":"line","source":"composite","source-layer":"admin","filter":["all",["==",["get","admin_level"],1],["==",["get","maritime"],"false"],["match",["get","worldview"],["all","US"],true,false]],"layout":{"line-join":"bevel"},"paint":{"line-color":"%236898B3","line-width":1.5,"line-dasharray":[1.5,1]}},
-    before_layer: 'road-label',
+   before_layer: 'road-label',
   });
 const staticImageUrl = request.url();
 // Now you can open staticImageUrl in a browser.
--- a/docs/services.md
+++ b/docs/services.md
@@ -509,7 +509,7 @@ const request = staticClient
     height: 300,
     position: 'auto',
     addlayer: {"id":"better-boundary","type":"line","source":"composite","source-layer":"admin","filter":["all",["==",["get","admin_level"],1],["==",["get","maritime"],"false"],["match",["get","worldview"],["all","US"],true,false]],"layout":{"line-join":"bevel"},"paint":{"line-color":"%236898B3","line-width":1.5,"line-dasharray":[1.5,1]}},
-    layer_id: 'road-label',
+    before_layer: 'road-label',
   });
 const staticImageUrl = request.url();
 // Now you can open staticImageUrl in a browser.
--- a/docs/services.md
+++ b/docs/services.md
@@ -391,6 +391,9 @@ SDK returned.
   - `config.highRes` **[boolean][158]**  (optional, default `false`)
   - `config.before_layer` **[string][151]?** The ID of the style layer
       that overlays should be inserted *before*.
+  - `config.addlayer` **[Object][150]?** Adds a Mapbox style layer to the map's style at render time. Can be combined with before_layer.
+  - `config.setfilter` **[Array][160]?** Applies a filter to an existing layer in a style using Mapbox's expression syntax. Must be used with layer_id.
+  - `config.layer_id` **[string][151]?** Denotes the layer in the style that the filter specified in setfilter is applied to.
   - `config.attribution` **[boolean][158]** Whether there is attribution
       on the map image. (optional, default `true`)
   - `config.logo` **[boolean][158]** Whether there is a Mapbox logo
@@ -475,6 +478,38 @@ const request = staticClient
       coordinates: [12, 13],
       zoom: 4
     }
+  });
+const staticImageUrl = request.url();
+// Now you can open staticImageUrl in a browser.
+```
+
+```javascript
+// Filter all buildings that have a height value that is less than 300 meters
+const request = staticClient
+  .getStaticImage({
+    ownerId: 'mapbox',
+    styleId: 'streets-v11',
+    width: 200,
+    height: 300,
+    position: 'auto',
+    setfilter: [">","height",300],
+    layer_id: 'building',
+  });
+const staticImageUrl = request.url();
+// Now you can open staticImageUrl in a browser.
+```
+
+```javascript
+// Paint all the state and province level boundaries associated with the US worldview with a dashed line and insert it below the road-label layer
+const request = staticClient
+  .getStaticImage({
+    ownerId: 'mapbox',
+    styleId: 'streets-v11',
+    width: 200,
+    height: 300,
+    position: 'auto',
+    addlayer: {"id":"better-boundary","type":"line","source":"composite","source-layer":"admin","filter":["all",["==",["get","admin_level"],1],["==",["get","maritime"],"false"],["match",["get","worldview"],["all","US"],true,false]],"layout":{"line-join":"bevel"},"paint":{"line-color":"%236898B3","line-width":1.5,"line-dasharray":[1.5,1]}},
+    layer_id: 'road-label',
   });
 const staticImageUrl = request.url();
 // Now you can open staticImageUrl in a browser.
--- a/docs/services.md
+++ b/docs/services.md
@@ -389,7 +389,7 @@ SDK returned.
       [`PathOverlay`][139],
       [`GeoJsonOverlay`][141]
   - `config.highRes` **[boolean][158]**  (optional, default `false`)
-  - `config.insertOverlayBeforeLayer` **[string][151]?** The ID of the style layer
+  - `config.before_layer` **[string][151]?** The ID of the style layer
       that overlays should be inserted *before*.
   - `config.attribution` **[boolean][158]** Whether there is attribution
       on the map image. (optional, default `true`)
--- a/lib/helpers/__tests__/url-utils.test.js
+++ b/lib/helpers/__tests__/url-utils.test.js
@@ -71,14 +71,6 @@ describe('appendQueryObject', () => {
     ).toBe('/foo/bar?test=one%2Ctwo');
   });
 
-  test('preserves array for setfilter', () => {
-    expect(
-      appendQueryObject('/foo/bar', {
-        setfilter: ['==', 'name_en', 'Canada']
-      })
-    ).toBe('/foo/bar?setfilter=["==","name_en","Canada"]');
-  });
-
   test('accepts objects', () => {
     expect(
       appendQueryObject('/foo/bar', {
--- a/lib/helpers/__tests__/url-utils.test.js
+++ b/lib/helpers/__tests__/url-utils.test.js
@@ -71,16 +71,6 @@ describe('appendQueryObject', () => {
     ).toBe('/foo/bar?test=one%2Ctwo');
   });
 
-  test('accepts objects', () => {
-    expect(
-      appendQueryObject('/foo/bar', {
-        test: { one: 'two', three: 'four' }
-      })
-    ).toBe(
-      '/foo/bar?test=%7B%22one%22%3A%22two%22%2C%22three%22%3A%22four%22%7D'
-    );
-  });
-
   test('skips undefined properties', () => {
     expect(
       appendQueryObject('/foo/bar', {
--- a/lib/helpers/__tests__/url-utils.test.js
+++ b/lib/helpers/__tests__/url-utils.test.js
@@ -71,6 +71,16 @@ describe('appendQueryObject', () => {
     ).toBe('/foo/bar?test=one%2Ctwo');
   });
 
+  test('accepts objects', () => {
+    expect(
+      appendQueryObject('/foo/bar', {
+        test: { one: 'two', three: 'four' }
+      })
+    ).toBe(
+      '/foo/bar?test=%7B%22one%22%3A%22two%22%2C%22three%22%3A%22four%22%7D'
+    );
+  });
+
   test('skips undefined properties', () => {
     expect(
       appendQueryObject('/foo/bar', {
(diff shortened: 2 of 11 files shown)


docs/services.md
```diff
@@ -480,6 +483,38 @@ const staticImageUrl = request.url();
...
+// Paint all the state and province level boundaries associated with the US worldview with a dashed line and insert it below the road-label layer
+const request = staticClient
+  .getStaticImage({
+    ownerId: 'mapbox',
+    styleId: 'streets-v11',
+    width: 200,
+    height: 300,
+    position: 'auto',
+    addlayer: {"id":"better-boundary","type":"line","source":"composite","source-layer":"admin","filter":["all",["==",["get","admin_level"],1],["==",["get","maritime"],"false"],["match",["get","worldview"],["all","US"],true,false]],"layout":{"line-join":"bevel"},"paint":{"line-color":"%236898B3","line-width":1.5,"line-dasharray":[1.5,1]}},
+    layer_id: 'road-label',
```
riastrad: `layer_id` is for `setfilter`, and I believe this example would actually return a 422 response from the API
```suggestion
    before_layer: 'road-label',
```

services/__tests__/static.test.js
```diff
@@ -416,4 +416,81 @@ describe('getStaticImage', () => {
...
+        source: 'composite',
+        'source-layer': 'building',
+        filter: [
+          'all',
+          ['>=', ['get', 'height'], 150],
+          ['match', ['get', 'underground'], ['false'], true, false]
+        ],
+        paint: { 'fill-color': '%235E8DFF', 'fill-opacity': 0.5 }
+      },
+      layer_id: 'tunnel-street-minor-low'
```
riastrad:
```suggestion
      before_layer: 'tunnel-street-minor-low'
```

services/__tests__/static.test.js
```diff
@@ -416,4 +416,81 @@ describe('getStaticImage', () => {
...
+          source: 'composite',
+          'source-layer': 'building',
+          filter: [
+            'all',
+            ['>=', ['get', 'height'], 150],
+            ['match', ['get', 'underground'], ['false'], true, false]
+          ],
+          paint: { 'fill-color': '%235E8DFF', 'fill-opacity': 0.5 }
+        },
+        layer_id: 'tunnel-street-minor-low'
```
riastrad:
```suggestion
        before_layer: 'tunnel-street-minor-low'
```

services/static.js
```diff
@@ -128,6 +131,37 @@ var Static = {};
...
+ * // Paint all the state and province level boundaries associated with the US worldview with a dashed line and insert it below the road-label layer
+ * const request = staticClient
+ *   .getStaticImage({
+ *     ownerId: 'mapbox',
+ *     styleId: 'streets-v11',
+ *     width: 200,
+ *     height: 300,
+ *     position: 'auto',
+ *     addlayer: {"id":"better-boundary","type":"line","source":"composite","source-layer":"admin","filter":["all",["==",["get","admin_level"],1],["==",["get","maritime"],"false"],["match",["get","worldview"],["all","US"],true,false]],"layout":{"line-join":"bevel"},"paint":{"line-color":"%236898B3","line-width":1.5,"line-dasharray":[1.5,1]}},
+ *     layer_id: 'road-label',
```
riastrad:
```suggestion
 *    before_layer: 'road-label',
```

services/static.js
```diff
@@ -128,6 +131,37 @@ var Static = {};
...
+ * @example
+ * // Filter all buildings that have a height value that is less than 300 meters
+ * const request = staticClient
+ *   .getStaticImage({
+ *     ownerId: 'mapbox',
+ *     styleId: 'streets-v11',
+ *     width: 200,
+ *     height: 300,
+ *     position: 'auto',
+ *     setfilter: [">","height",300],
```
riastrad: It's not currently possible to use the `'auto'` parameter with style parameters (you will receive a `422` response from the API), _unless_ you are also using an overlay. So I would recommend using an actual `0,0,0` type position for all of these examples.

services/static.js
```diff
@@ -128,6 +131,37 @@ var Static = {};
...
+ *
+ * @example
+ * // Paint all the state and province level boundaries associated with the US worldview with a dashed line and insert it below the road-label layer
+ * const request = staticClient
+ *   .getStaticImage({
+ *     ownerId: 'mapbox',
+ *     styleId: 'streets-v11',
+ *     width: 200,
+ *     height: 300,
+ *     position: 'auto',
```
riastrad: Same comment as my previous one!

services/static.js
```diff
@@ -182,8 +219,21 @@ Static.getStaticImage = function(config) {
...
+  }
+  if (config.setfilter !== undefined) {
+    query.setfilter = config.setfilter;
+  }
+  if (config.layer_id !== undefined) {
+    query.layer_id = config.layer_id;
+  }
+
+  if (config.setfilter !== undefined && config.layer_id === undefined) {
+    throw new Error('setfilter requires layer_id');
```
riastrad: This section is 💯, I would also add a check to make sure that `setfilter` and `addlayer` aren't being used simultaneously (as this is not currently supported by the API)

services/static.js
katydecorah: Thank you! This will be another good test!

services/static.js
katydecorah: Thank you! I'll also comb through the error messages to make sure everything is covered! https://docs.mapbox.com/api/maps/#static-images-api-errors

services/__tests__/static.test.js
```diff
@@ -416,4 +416,163 @@ describe('getStaticImage', () => {
...
+            ['match', ['get', 'underground'], ['false'], true, false]
+          ],
+          paint: { 'fill-color': '%235E8DFF', 'fill-opacity': 0.5 }
+        },
+        before_layer: 'tunnel-street-minor-low',
+        setfilter: ['in', 'code', 'CA'],
+        layer_id: 'tunnel-street-minor-low'
+      });
+    }).toThrow(/addlayer and setfilter cannot be used in the same request/);
+  });
```
riastrad: The test itself is correct, but the description is not 😬 

You can't use `setfilter` and `addlayer` together _even if_ you're using an overlay.

katydecorah: 🤦‍♂ thanks for catching that!

services/__tests__/static.test.js
```diff
@@ -416,4 +416,163 @@ describe('getStaticImage', () => {
...
+          paint: { 'fill-color': '%235E8DFF', 'fill-opacity': 0.5 }
+        },
+        before_layer: 'tunnel-street-minor-low'
+      });
+    }).toThrow(
+      /Auto extent cannot be used with style parameters and no overlay/
+    );
+  });
+
+  test('addlayer and setfilter cannot be used together unless also using an overlay', () => {
```
katydecorah:
```suggestion
  test('addlayer and setfilter cannot be used in the same request', () => {
```

services/static.js
```diff
@@ -42,8 +42,11 @@ var Static = {};
  *   [`GeoJsonOverlay`](#geojsonoverlay)
  *
  * @param {boolean} [config.highRes=false]
- * @param {string} [config.insertOverlayBeforeLayer] - The ID of the style layer
+ * @param {string} [config.before_layer] - The ID of the style layer
  *   that overlays should be inserted *before*.
+ * @param {Object} [config.addlayer] - Adds a Mapbox style layer to the map's style at render time. Can be combined with before_layer.
+ * @param {Array} [config.setfilter] - Applies a filter to an existing layer in a style using Mapbox's expression syntax. Must be used with layer_id.
```
danswick: I ran into some issues while testing this locally. The way we parse query params expects that any array is just a list of things you want to append as new parameters, so we just turn those arrays into a string of comma-separated values. This turns something like `[">", "height", 300]` into `>,height,300`, which makes the static API unhappy because it expects an actual array.

danswick: @katydecorah i realized my above suggestion wasn't working, so I pushed a couple of other changes just for the sake of discussion. I'm sure there's a cleaner way to do this! https://github.com/mapbox/mapbox-sdk-js/pull/330/commits/85fd40f34c61d856898faa27b34489d13073c69f

katydecorah: > I ran into some issues while testing this locally.

Do you have any example of a query that you ran into issues with locally?

danswick: @katydecorah I just grabbed some examples from the tests: 

```
const staticReq = staticService.getStaticImage({
  ownerId: 'mapbox',
  styleId: 'streets-v11',
  width: 500,
  height: 600,
  position: {
    coordinates: [-74.01254,40.70577],
    zoom: 16
  },
  setfilter: [">","height",100],
  layer_id: 'building'
});
```

Which resulted in a request URL of:

```
https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/-74.01254,40.70577,16/500x600?setfilter=%3E%2Cheight%2C100&layer_id=building&access_token=
```

katydecorah: Ohh I see! Hmm yes, v funky

lib/helpers/url-utils.js
```diff
@@ -29,8 +29,15 @@ function appendQueryParam(url, key, value) {
...
-  if (value !== undefined && value !== '' && value !== true) {
+  if (
+    value !== undefined &&
+    value !== '' &&
+    value !== true &&
+    key !== 'setfilter'
+  ) {
     query += '=' + encodeValue(value);
+  } else {
+    query += '=' + value;
```
katydecorah: Looks like this code will preserve `setfilter` value so it doesn't get encoded. Right now `addlayer` is being encoded. The encoded versions and decoded versions both work. 

We should decide if both values should be encoded or left decoded to make it consistent. It seems like the default is to *encode* which I'm assuming is how all other APIs in this library are being handled by this function - @danswick do you see any reason why we shouldn't stick with encoding for these values? 

I think the decoded values are much easier to read and take up fewer characters. Although encoding characters seems to be a good practice, plus the developer could wrap `decodeURIComponent` around `url()` if they desired a prettier URL (which is what I'm currently doing as I'm developing with these new params for the static playground).

My vote leans towards encoding.


Examples:

`addlayer` - encoded (current state):

```
https://api.mapbox.com/styles/v1/mapbox/light-v10/static/-77.03968,38.89744,5/300x200?addlayer=%7B%22id%22%3A%22road-overlay%22%2C%22type%22%3A%22line%22%2C%22source%22%3A%22composite%22%2C%22source-layer%22%3A%22road%22%2C%22filter%22%3A%5B%22%3D%3D%22%2C%5B%22get%22%2C%22class%22%5D%2C%22motorway%22%5D%2C%22paint%22%3A%7B%22line-color%22%3A%22%23ff0000%22%2C%22line-width%22%3A5%7D%7D&access_token=
```

`addlayer` - decoded:
```
https://api.mapbox.com/styles/v1/mapbox/light-v10/static/-77.03968,38.89744,5/300x200?addlayer={"id":"road-overlay","type":"line","source":"composite","source-layer":"road","filter":["==",["get","class"],"motorway"],"paint":{"line-color":"%23ff0000","line-width":5}}&access_token=
```

`setfilter` - encoded: 

```
https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/-91,60,2/800x600?setfilter=%5B%22%3D%3D%22%2C%22name_en%22%2C%22Canada%22%5D&layer_id=country-label&access_token=
```

`setfilter` - decoded: 

```
https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/-91,60,2/800x600?setfilter=["==","name_en","Canada"]&layer_id=country-label&access_token=
```

danswick: Really good point! i also lean toward encoding. Coincidentally, I had a dream about this overnight and realized that all of the changes I made yesterday are unnecessary 😂. I reverted all of my URL utils changes and changed the static service to stringify `setFilter` values instead, which I think solves both problems https://github.com/mapbox/mapbox-sdk-js/pull/330/files#diff-a033a8610015f7c46139070fceb2ff39R235.

lib/helpers/__tests__/url-utils.test.js
```diff
@@ -71,6 +71,24 @@ describe('appendQueryObject', () => {
     ).toBe('/foo/bar?test=one%2Ctwo');
   });
 
+  test('preserves array for setfilter', () => {
```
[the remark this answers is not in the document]
katydecorah: I think we'll want to keep this test to confirm that setfilter maintains its array shape in the url since the default condition is to turn it into a comma-delineated list

services/static.js
```diff
@@ -232,7 +232,7 @@ Static.getStaticImage = function(config) {
     query.addlayer = config.addlayer;
   }
   if (config.setfilter !== undefined) {
-    query.setfilter = JSON.stringify(config.setfilter);
+    query.setfilter = config.setfilter;
```
katydecorah: 📝This value will get stringify'd at the appendQueryObject() function

lib/helpers/url-utils.js
```diff
@@ -54,7 +54,7 @@ function appendQueryObject(url, queryObject) {
     if (value === undefined) {
       return;
     }
-    if (Array.isArray(value)) {
+    if (Array.isArray(value) && key !== 'setfilter') {
```
katydecorah: 📝I think we need to keep this to make sure `setfilter` retains its array shape

lib/helpers/__tests__/url-utils.test.js
```diff
@@ -76,7 +76,9 @@ describe('appendQueryObject', () => {
       appendQueryObject('/foo/bar', {
         setfilter: ['==', 'name_en', 'Canada']
       })
-    ).toBe('/foo/bar?setfilter=["==","name_en","Canada"]');
+    ).toBe(
+      '/foo/bar?setfilter=%5B%22%3D%3D%22%2C%22name_en%22%2C%22Canada%22%5D'
```
katydecorah: 📝This test asserts that `setfilter` retains its array shape and is encoded

lib/helpers/url-utils.js
danswick: Sorry for making things extra confusing, but I think it is a bit cleaner to stringify in the static service and leave the special case out of `url-utils`. That way we can keep `url-utils` generalized and protect ourselves from dealing with more service-specific edge cases in the future. 

What do you think?

lib/helpers/__tests__/url-utils.test.js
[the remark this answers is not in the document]
katydecorah: I was mixed up in my previous comment, you're correct. Removed this test.

lib/helpers/url-utils.js
```diff
@@ -61,6 +61,9 @@ function appendQueryObject(url, queryObject) {
         })
         .join(',');
     }
+    if (typeof value === 'object') {
+      value = JSON.stringify(value);
```
katydecorah: Maybe we should stringify addlayer the same way so we don't have to change this function at all!

PR #341, opened 2019-11-19
Correct `listTilesets` docs

chrisbrainerd:
Docs omitted a `.send()` which is necessary

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -998,6 +998,7 @@ List a user's tilesets.
 
 ```javascript
 tilesetsClient.listTilesets()
+  .send()
   .then(response => {
     const tilesets = response.body;
   });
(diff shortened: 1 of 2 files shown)


chrisbrainerd:
@andrewharvey can do, although looks like there's a whole lot of formatting that'll be included. Any chance someone changed some settings since that was run last? I tried running with my local linter/prettier off thinking maybe somehow it was using those settings and still I'm ending up with a _massive_ diff, entirely just formatting things.

andrewharvey:
> can do, although looks like there's a whole lot of formatting that'll be included. Any chance someone changed some settings since that was run last? I tried running with my local linter/prettier off thinking maybe somehow it was using those settings and still I'm ending up with a massive diff, entirely just formatting things.

I wouldn't worry too much since it's a generated file. In an ideal world these generated files wouldn't be committed, but we don't have the same level of docs infrastructure here compared to say the GL JS docs.

PR #345, opened 2019-12-13
feature(tilesets): List queries, tileJSON metadata

chooban:
Add the query parameters for listing tilesets, and a new service command
for retrieving TileJSON metadata.

### DIFF
--- a/services/__tests__/tilesets.test.js
+++ b/services/__tests__/tilesets.test.js
@@ -27,3 +27,14 @@ describe('listTilesets', () => {
     });
   });
 });
+
+describe('tileJSONMetadata', () => {
+  test('works', () => {
+    tilesets.tileJSONMetadata(undefined, { tilesetId: 'hello-world' });
+    expect(tu.requestConfig(tilesets)).toEqual({
+      path: '/v4/:tilesetId.json',
+      method: 'GET',
+      params: { tilesetId: 'hello-world' }
+    });
+  });
+});
--- a/services/__tests__/tilesets.test.js
+++ b/services/__tests__/tilesets.test.js
@@ -10,11 +10,11 @@ beforeEach(() => {
 
 describe('listTilesets', () => {
   test('works', () => {
-    tilesets.listTilesets({ ownerId: 'specialguy' });
+    tilesets.listTilesets();
     expect(tu.requestConfig(tilesets)).toEqual({
       path: '/tilesets/v1/:ownerId',
       method: 'GET',
-      params: { ownerId: 'specialguy' },
+      params: {},
       query: {}
     });
   });
--- a/services/__tests__/tilesets.test.js
+++ b/services/__tests__/tilesets.test.js
@@ -10,16 +10,6 @@ beforeEach(() => {
 
 describe('listTilesets', () => {
   test('works', () => {
-    tilesets.listTilesets();
-    expect(tu.requestConfig(tilesets)).toEqual({
-      path: '/tilesets/v1/:ownerId',
-      method: 'GET',
-      params: {},
-      query: {}
-    });
-  });
-
-  test('works with specified ownerId', () => {
     tilesets.listTilesets({ ownerId: 'specialguy' });
     expect(tu.requestConfig(tilesets)).toEqual({
       path: '/tilesets/v1/:ownerId',
--- a/services/__tests__/tilesets.test.js
+++ b/services/__tests__/tilesets.test.js
@@ -14,7 +14,8 @@ describe('listTilesets', () => {
     expect(tu.requestConfig(tilesets)).toEqual({
       path: '/tilesets/v1/:ownerId',
       method: 'GET',
-      params: undefined
+      params: {},
+      query: {}
     });
   });
 
@@ -23,14 +24,29 @@ describe('listTilesets', () => {
     expect(tu.requestConfig(tilesets)).toEqual({
       path: '/tilesets/v1/:ownerId',
       method: 'GET',
-      params: { ownerId: 'specialguy' }
+      params: { ownerId: 'specialguy' },
+      query: {}
+    });
+  });
+
+  test('works with query params', () => {
+    tilesets.listTilesets({
+      ownerId: 'specialguy',
+      limit: 250,
+      type: 'raster'
+    });
+    expect(tu.requestConfig(tilesets)).toEqual({
+      path: '/tilesets/v1/:ownerId',
+      method: 'GET',
+      params: { ownerId: 'specialguy' },
+      query: { limit: 250, type: 'raster' }
     });
   });
 });
 
 describe('tileJSONMetadata', () => {
   test('works', () => {
-    tilesets.tileJSONMetadata(undefined, { tilesetId: 'hello-world' });
+    tilesets.tileJSONMetadata({ tilesetId: 'hello-world' });
     expect(tu.requestConfig(tilesets)).toEqual({
       path: '/v4/:tilesetId.json',
       method: 'GET',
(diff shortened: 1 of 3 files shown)


services/tilesets.js
```diff
@@ -16,6 +16,11 @@ var Tilesets = {};
  *
  * @param {Object} [config]
  * @param {string} [config.ownerId]
+ * @param {Object} [query]
+ * @param {number} [query.range]
```
andrewharvey: As far as I can see `range` is not defined https://docs.mapbox.com/api/maps/#list-tilesets

services/tilesets.js
```diff
@@ -16,6 +16,11 @@ var Tilesets = {};
...
+ * @param {string} [query.sortBy]
+ * @param {string}  [query.start]
```
andrewharvey: In keeping with other parts of this library, like https://github.com/mapbox/mapbox-sdk-js/blob/master/services/styles.js#L175 let's include a description of these fields here.

chooban:
No problem! Those should be updated now. I also rebased the PR onto current master.

services/tilesets.js
```diff
@@ -16,6 +16,12 @@ var Tilesets = {};
  *
  * @param {Object} [config]
  * @param {string} [config.ownerId]
+ * @param {Object} [query]
```
andrewharvey: After another look, I think it's better to be consistent with what all the other methods do and that's use config for everything and then pick the attributes for params and query. For example see https://github.com/mapbox/mapbox-sdk-js/blob/master/services/datasets.js#L180-L209.

services/tilesets.js
```diff
@@ -16,6 +16,12 @@ var Tilesets = {};
...
+ * @param {string} [query.type] - Filter results by tileset type, either `raster` or `vector`.
```
andrewharvey: how about making the jsdoc an enum of values like https://github.com/mapbox/mapbox-sdk-js/blob/master/services/geocoding.js#L37 ?

services/tilesets.js
```diff
@@ -16,6 +16,12 @@ var Tilesets = {};
...
+ * @param {number} [query.limit] - The maximum number of tilesets to return, from 1 to 500. The default is 100.
```
andrewharvey: you can specify the default within the jsdoc like https://github.com/mapbox/mapbox-sdk-js/blob/master/services/geocoding.js#L44

services/tilesets.js
```diff
@@ -33,13 +39,45 @@ var Tilesets = {};
...
+    start: v.string,
+    type: v.oneOf('raster', 'vector'),
+    visibility: v.oneOf('public', 'private')
   })(config);
 
   return this.client.createRequest({
     method: 'GET',
     path: '/tilesets/v1/:ownerId',
-    params: config
+    params: pick(config || {}, ['ownerId']),
```
chooban: This isn't right! Or the line below!

If I don't have a default object then I get a lot of failing tests which seem to be to do with pagination. I can't see what I'm doing differently here when compared to something like the datasets service.

services/tilesets.js
chooban: Not sure why I decided to blaze my own trail with that one. :/

services/tilesets.js
danswick: Oh hm, this is kind of a weird one because there are no *required* query params. There are a couple of things we could do: 

1. `ownerId` is required, so we'll want to update the `'works'` test to include it (I think this changed since the service was initially added to `sdk-js`).
2. Since there are no required (or even publicly documented 😂) query params for the list tilesets endpoint, there do seem to be legitimate situations where there will be no `query`s to `pick`. Your approach seems pretty solid to me, but I might change it to something like `config ? pick(config, ['blah', 'bluh']) : {}` to make it a little clearer what's happening. Mostly personal preference, but I think this would also prevent some unnecessary `pick` calls.

services/tilesets.js
```diff
@@ -16,6 +17,11 @@ var Tilesets = {};
  *
  * @param {Object} [config]
  * @param {string} [config.ownerId]
+ * @param {'raster'|'vector'} [config.type] - Filter results by tileset type, either `raster` or `vector`.
+ * @param {number} [config.limit=100] - The maximum number of tilesets to return, from 1 to 500. The default is 100.
```
andrewharvey: great now that the default is specified with JSDoc there's no need to include that in the description as you'll see in the generated docs/services.md it get's appended.

andrewharvey:
Looks like we're almost there @chooban. I've started working on the rest of the tilesets api https://github.com/mapbox/mapbox-sdk-js/issues/353.

services/tilesets.js
chooban: Makes sense. I've updated the `works` test and the picking of config params. I went with the ternary as I agree that it looks a little clearer than having the logic inside the function parameters.

services/__tests__/tilesets.test.js
```diff
@@ -10,20 +10,37 @@ beforeEach(() => {
 
 describe('listTilesets', () => {
   test('works', () => {
-    tilesets.listTilesets();
+    tilesets.listTilesets({ ownerId: 'specialguy' });
```
andrewharvey: originally we had two unit tests, one without ownerId one with, any reason we can't keep both tests?

chooban: Going by @danswick's comments, I thought it might not be valid to have it any more but I'm not sure. 

So if we have a test without `ownerId` that just shows we have an empty params and query object in the result, is that what we'd want from a `works` test? Or should `works` be the one with an `ownerId`? 

Should I update the validation in listTilesets to use `ownerId: v.required(v.string)`?

andrewharvey: ownerId should be optional here because although the API requires it https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/classes/mapi-request.js#L127 set's it based on the token if not supplied.

We should have two unit tests here to make sure both omitting and supplying it works.

services/tilesets.js
andrewharvey: @danswick Is ownerId required in this method? https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/classes/mapi-request.js#L127 set's it based on the access token if not supplied.

services/__tests__/tilesets.test.js
chooban: Test reinstated. There has been a change from an `undefined` value when there are no params to an empty object, which would be trivial to change in the `tilesets` service. 

Apologies for there being so many things to change in such a small PR!

danswick: Ack! Good point, @andrewharvey. And I apologize for leading you astray, @chooban. Poking around a bit more, I don't believe this endpoint supports listing tilesets for accounts other than the one associated with the token provided. I'm double checking with the team who works on that API to be extra sure, but am consistently getting 404 responses from the `list tilesets` endpoint, even for accounts I know have public tilesets. So we may still want to omit it as a param, but not for the reason I originally gave you.

chooban: No problem at all! Hopefully I can take what I've learned in this PR and add a few more parts of the API.

So at the moment we've got a unit test with no ownerId, and one with. Do we pause until there's an answer from the API team?

andrewharvey: Let's see what @danswick says but I think we should go ahead with this PR leaving ownerId unaffected, that is unchanged from the current behaviour, with the intention to keep the PR small and focused on the original intent. and if it turns out we shouldn't expose ownerId as an option at all, then we can follow up with a separate PR.

danswick: Great suggestion, @andrewharvey. I opened an [issue](https://github.com/mapbox/mapbox-sdk-js/issues/355) to track this at a later time. Let's go ahead here!

chooban:
You're welcome!

PR #348, opened 2019-12-18
add support for draft styles

andrewharvey:
- [x] unit tests
- [x] real world tests to ensure I've covered every endpoints that supports draft, and none that don't actually support it.

### DIFF
--- a/services/styles.js
+++ b/services/styles.js
@@ -300,7 +300,7 @@ Styles.deleteStyleIcon = function(config) {
  * @param {boolean} [config.highRes] - If true, returns spritesheet with 2x
  *   resolution.
  * @param {string} [config.ownerId]
- * @param {boolean} [config.draft=false] If `true` will retrieve the draft style sprite, otherwise will retrieve the published style sprite.
+ * @param {boolean} [config.draft=false] - If `true` will retrieve the draft style sprite, otherwise will retrieve the published style sprite.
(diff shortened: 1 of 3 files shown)


services/styles.js
```diff
@@ -368,22 +380,24 @@ Styles.getFontGlyphRange = function(config) {
...
     styleId: v.required(v.string),
     scrollZoom: v.boolean,
     title: v.boolean,
-    ownerId: v.string
+    ownerId: v.string,
+    draft: v.boolean
   })(config);
 
-  var fileName = config.styleId + '.html';
+  var fileName = config.styleId + (config.draft ? '/draft' : '') + '.html';
```
danswick: When I was testing locally, the `/` in `/draft` was getting URI encoded. Do you think this is another case like https://github.com/mapbox/mapbox-sdk-js/pull/352#issuecomment-578498228?

andrewharvey: oh good catch, you're right.

andrewharvey:
I've resolved the merge conflicts and rebased after #356.

It would be appreciated if this could get some :eyes: to review, if it can be merged into master it would help prevent extra work resolving conflicts :wink:

andrewharvey:
I'm not sure who's the maintainer of this project so I've requested a review from all those mentioned in https://github.com/mapbox/mapbox-sdk-js/pull/356

PR #351, opened 2020-01-26
set encoding as binary for png requests

### LINKED ISSUE #317: Save Mapbox static api response to png file with node sdk js
I have the same problem as this Stackoverflow question I found which I'm reproducing for your benefit here:

https://stackoverflow.com/questions/54117885/save-mapbox-static-api-response-to-png-file-with-node-sdk-js

The png file created with the following code using mapbox-sdk-js is unreadable, yet the response headers display content "image/png" and status code is 200. This is not a problem with my token or else, if I use curl with the same parameters I get a proper PNG file (and a nice map).

Could you help me to save the response to a file ? Thanks !

```javascript
const mbxStatic = require('@mapbox/mapbox-sdk/services/static')
const fs = require('fs')

// Get config
const MAPBOX_TOKEN = process.env.MAPBOX_TOKEN

const staticClient = mbxStatic({ accessToken: MAPBOX_TOKEN })

staticClient
  .getStaticImage({
    ownerId: 'mapbox',
    styleId: 'streets-v10',
    width: 200,
    height: 300,
    position: {
      coordinates: [-77, 38],
      zoom: 5
    }
  })
  .send()
  .then(response => {
    const image = response.body
    console.log(response.headers)
    fs.writeFileSync("./map.png", image, 'binary')
  }).catch(error => console.log(error));
```

### LINKED ISSUE #342: Static image not returning binary data/string
I'm saving the body using Buffer.from(body, 'binary') but what I'm getting out isn't valid.

```
for(let i = 0; i < image.length; i++)
    if(image.charCodeAt(i) > 255) {
        console.log([image.charCodeAt(i), i, image[i]]);
            break;
    }
```

```[ 65533, 0, '�' ]```

65533 is javascriptese for dunno which means it tried to load as UTF8 but wasn't.

```
// First char of PNG magic.
'\x89'.charCodeAt(0)
'\u0089'.charCodeAt(0)
```

I get the same with rawBody.

Is there anyway to control this? I see perhaps some issues making the lib work on browser which might be why it's not using buffer but also even for browser if it's not using binary strings then it'll be broken there too. It seems to only be useful if you generate the URL and use that?

Is these a missing layer for conversion in the lib, IE to expect binary, JSON, etc?

I see just this:

```
  try {
    this.body = JSON.parse(responseData.body || '{}');
  } catch (parseError) {
    this.body = responseData.body;
  }
```

Technically speaking binary data can accidentally parse as JSON.

This needs an implementation for string and buffer:
```
      var body = '';
      gotStream.on('data', function(chunk) {
        body += chunk;
      });
```

And this needs to be configurable:

```
gotStream.setEncoding('utf8');
```

Which would pass down to the stream consumer used.

I don't know what happens with the browser layer but small chance I reckon that it returns binary string where as this has explicitely utf8 so will break on non-utf8.

andrewharvey:
closes #317 
closes #342

### DIFF
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -62,6 +62,7 @@ The `emitter` property is an `EventEmitter` that emits the following events:
 - `file` **([Blob][25] \| [ArrayBuffer][26] \| [string][23] | ReadStream)** A file to
     send with the request. The browser client accepts Blobs and ArrayBuffers;
     the Node client accepts strings (filepaths) and ReadStreams.
+- `encoding` **[string][23]** The encoding of the response.
 
 ### url
 
--- a/docs/services.md
+++ b/docs/services.md
@@ -308,6 +308,17 @@ stylesClient.getStyleSprite({
   .then(response => {
     const sprite = response.body;
   });
+
+stylesClient.getStyleSprite({
+  format: 'png',
+  styleId: 'foo',
+  highRes: true
+})
+  .send()
+  .then(response => {
+    const sprite = response.body;
+    fs.writeFileSync('sprite.png', sprite, 'binary');
+  });
 ```
 
 Returns **MapiRequest** 
--- a/docs/services.md
+++ b/docs/services.md
@@ -308,7 +308,9 @@ stylesClient.getStyleSprite({
   .then(response => {
     const sprite = response.body;
   });
+```
 
+```javascript
 stylesClient.getStyleSprite({
   format: 'png',
   styleId: 'foo',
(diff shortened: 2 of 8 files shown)


andrewharvey:
resolved conflicts and rebased against the new master

andrewharvey:
Thanks for merging @tristen it makes my life so much easier. I've been going crazy trying to maintain my own fork of mapbox-sdk-js, keeping 4 conflicting branches in sync topped off by yarn not always updating correctly when pointing to a github branch.

PR #352, opened 2020-01-26
fix sprite highres encoding

andrewharvey:
previously requests for highres sprites were failing due to `sprite@2x.png` being url encoded as `%40` which fails when requested via the Mapbox API.

### DIFF
--- a/lib/helpers/url-utils.js
+++ b/lib/helpers/url-utils.js
@@ -107,10 +107,7 @@ function interpolateRouteParams(route, params) {
     if (value === undefined) {
       throw new Error('Unspecified route parameter ' + paramId);
     }
-    var preppedValue =
-      typeof value === 'string' && value.startsWith('sprite@')
-        ? value
-        : encodeValue(value);
+    var preppedValue = encodeValue(value);
     return '/' + preppedValue;
   });
 }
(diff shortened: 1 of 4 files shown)


andrewharvey:
alternativly this could be implemented in the same way as https://github.com/mapbox/mapbox-sdk-js/blob/3b7ba16925f745b5db0f5268637cbcefe9c3cb67/services/static.js#L263 setting a `preEncodedUrlParts`, @davidtheclark any preference which approach to use?

danswick:
Thanks for this PR, @andrewharvey! I think I have a slight preference for the method you link to in the static service. `@` characters appear in a few other cases as well and I feel like it might be better to handle them on a service-by-service basis. What do you think?

andrewharvey:
Either way it's an ugly hack but I'm inclined to agree with you it's better to deal with each case in the service rather than adding service specific code to url-utils.

andrewharvey:
manually resolved conflicts and rebased against the latest master

andrewharvey:
manually resolved conflicts and rebased against the latest master

andrewharvey:
forced pushed to get travis to retry since it errored with

`fatal: unable to access 'https://github.com/mapbox/mapbox-sdk-js.git/': Failed to connect to github.com port 443: Connection timed out`

That happens sporadically, best we can do is just retry.

tristen:
👋@danswick pinging you to review this one again

PR #356, opened 2020-01-30
Support a metadata param when creating a style

tristen:
Required for applications relying on writing a style out with the
metadata object after the Styles API begins stripping that data out for
performance.


andrewharvey:
Will the API documentation at https://docs.mapbox.com/api/maps/#create-a-style also be updated to indicate this parameter?

elifitch:
@andrewharvey Sure is! A PR is in-flight and will go out at the same time

elifitch:
@lednax this is now ready for review

PR #358, opened 2020-02-08
prepare 0.10.0

andrewharvey:
Preparing a new release, there's been quite a few changes since the last release and it would be good to get these published.

The only concern I have is the two new methods `tilesetStatus` and `tilesetJob` should they actually be `getTilesetStatus` and `getTilesetJob`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - **Fix:** Fix responses containing binary data when using `Static#getStaticImage`, `Styles#getStyleSprite` and `Styles#getFontGlyphRange`.
 - **Fix:** Fix requests for highres sprites in `Styles#getStyleSprite`.
 - **Fix:** set `position.bearing` to `0` if `position.pitch` is defined and `position.bearing` is not in the Static API.
+- **Fix:** use `tilesets.getRecipe` in tilesets API example.
 
 ## 0.9.0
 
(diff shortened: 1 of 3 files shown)


danswick:
Thanks for the prep PR, @andrewharvey! Before we get a release out, I would like to do some more testing with the new Tilesets API endpoints and options. @katydecorah and I had a couple of internal questions about whether we need to take any special considerations for endpoints that are still in beta.

katydecorah:
Noting that it would be great to include #360 in this release too

andrewharvey:
rebased to include #360 

I hadn't realised this repo uses npm instead of yarn, so I also installed with npm and re-generated docs to ensure they are in sync.

andrewharvey:
Thanks @danswick I tagged the commit and pushed it. I couldn't find release documentation so I'll put something together. Specifically all the old releases pre the move to the `@mapbox` namespace are in the form v0.10.0 but post `@mapbox` namespace are `0.10.0`. Although there is no conflict in npm due to one being namespaced and the other not, there is potential in git, so seems like we're stuck with this convention of release tag naming.

danswick:
Oh that's a good point about release documentation, @andrewharvey. Since we're in the `@mapbox` namespace now, someone from Mapbox will need to publish to NPM. I can add that to the README, along with a note about versioning convention.

andrewharvey:
> Oh that's a good point about release documentation, @andrewharvey. Since we're in the @mapbox namespace now, someone from Mapbox will need to publish to NPM. I can add that to the README, along with a note about versioning convention.

That would be great if you want to do it. :+1:

Issue #315, opened 2019-04-22
Static API service `position.bearing` not added to url


chrisgo:
Using example code here
https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/services.md#static

```
const request = staticClient
  .getStaticImage({
    ownerId: 'mapbox',
    styleId: 'streets-v11',
    width: 600,
    height: 400,
    position: {
      coordinates: [-122.4241, 37.78],
      zoom: 14,
      bearing: 0,
      pitch: 60,
    }
  });
const staticImageUrl = request.url();
```

The expected URL is

https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/-122.4241, 37.78,14,0,60/600x450/600x400

but it is coming out as

https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/-122.4241, 37.78,14,60/600x450/600x400

It is missing the `bearing` in the URL output of `encodePosition()`

The expected URL has an extra `0,` between `14,` and `60`

The code on this line
https://github.com/mapbox/mapbox-sdk-js/blob/master/services/static.js#L202

```
.filter(Boolean)
```
seems to be removing `0` due to a Boolean filter

kepta:
@chrisgo happy to review a PR fixing this.

PR #360, opened 2020-02-20
If position.pitch is set then position.bearing must be 0

katydecorah:
Right now if you set the pitch, but not bearing in the Static API, the request does not put a 0 in for bearing. As a result the SDK sets the bearing and not the pitch.

Steps to reproduce:

1. Visit https://docs.mapbox.com/playground/static/
2. Keep bearing at `0` and set pitch to `60`
3. See the request URL, where the pitch is set in bearing's place:

```
https://api.mapbox.com/styles/v1/mapbox/light-v10/static/-77.0397,38.8974,5,60/300x200?access_token=
```

Expected:

```
https://api.mapbox.com/styles/v1/mapbox/light-v10/static/-77.0397,38.8974,5,0,60/300x200?access_token=
```

Fixes https://github.com/mapbox/mapbox-sdk-js/issues/315

### DIFF
--- a/services/static.js
+++ b/services/static.js
@@ -273,7 +273,7 @@ function encodePosition(position) {
   return position.coordinates
     .concat([
       position.zoom,
-      position.pitch && !position.bearing ? 0 : position.bearing, // if pitch is set, bearing must be 0
+      position.pitch && !position.bearing ? 0 : position.bearing, // if pitch is set, but bearing is not, bearing must be 0
       position.pitch === 0 ? undefined : position.pitch
     ])
     .filter(function(el) {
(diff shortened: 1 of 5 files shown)


services/static.js
```diff
@@ -271,8 +271,14 @@ function encodePosition(position) {
   if (position === 'auto') return 'auto';
 
   return position.coordinates
-    .concat([position.zoom, position.bearing, position.pitch])
-    .filter(Boolean)
+    .concat([
+      position.zoom,
+      position.pitch && !position.bearing ? 0 : position.bearing, // if pitch is set, bearing must be 0
```
danswick: Just small suggestion for the benefit of our future selves!

```suggestion
      position.pitch && !position.bearing ? 0 : position.bearing, // if pitch is set, but bearing is not, bearing must be 0
```

Issue #213, opened 2018-04-23
TypeScript definitions for mapbox-sdk-js


ctaggart:
I have not been able to find TypeScript definitions for this repository. Are there any? If not, please consider this a feature request.

kepta:
Happy to take a pull request for typescript types.

m00dy:
PLEASE PLEASE PLEASE :)

AndyCJ:
I've done the typings for the MapiClient, and the Geocoder Service.

 - They aren't complete.
 - They aren't very well tested.
 - They don't have documentation.

But, they can be a starting point for anyone else who's interested...

Hopefully someone finds them useful.

<details>
	<summary>mapbox-sdk.d.ts</summary>

    declare module "@mapbox/mapbox-sdk" {
        // https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/services.md#coordinates
        export type Coordinates = [number, number];
    
        // https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/services.md#boundingbox
        export type BoundingBox = [number, number, number, number];
    
        // https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/classes.md#mapirequest
        export type Emitter = 'response' | 'error' | 'downloadProgress' | 'uploadProgress';
    
        export interface MapiRequest {
            emitter: Emitter;
            client: MapiClient;
    
            response: MapiResponse | null;
            error: MapiError | Error | null;
    
            aborted: boolean;
            sent: boolean;
    
            path: string;
            origin: string;
            method: string;
            query: Object;
            params: Object;
            headers: Object;
    
            body?: Object | string;
            file?: Blob | ArrayBuffer | string;
    
            url: (accessToken?: string) => string;
            send: () => Promise<MapiResponse>;
            abort: () => void;
            eachPage: (callback: EachPageCallback) => void;
            clone(): MapiRequest;
        }
    
        // https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/classes.md#eachpage
        export interface EachPageCallback {
            (error: MapiError, response: MapiResponse, next: () => void): void;
        }
    
        // https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/classes.md#mapiresponse
        export interface MapiResponse<TBody extends Object = Object> {
            body: TBody;
            rawBody: string;
            statusCode: number;
            headers: Object;
            links: Object;
            request: MapiRequest;
    
            hasNextPage(): boolean;
[...]

PR #361, opened 2020-03-16
Use jsdoc to generate a typescript stub

SKalt:
I added an npm script, `typecheck`, which checks the type annotations in the JSDoc comments.  I wrangled it down to two errors caused by non-es6-class prototypical inheritance, but couldn't get the type checking any better. Regardless of exit code, `npm run typecheck` generates `./index.d.ts`, which addresses #213  .

Implementation note: you'll notice 
```js
/** @ignore @typedef {import("./path/to/module")} ImportedType */
```
scattered throughout the code. The `@ignore` is to prevent duplicate entries in the markdown docs.

If I refactored the inheritance code from old-style 
```
function MyClass(){ this.prop = value};
MyClass.prototype = Object.create(OtherClass.prototype)
```
to
```
class MyClass extends OtherClass {
  constructor() { this.prop = value }
}
```
I'm positive I could eliminate the last of the type errors. 

Reference: 
- https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html


SKalt:
In 97e6809, I got `npm run typecheck` to pass at the expense of IE11 support.

andrewharvey:
This seems okay to me, people like myself who don't use typescript probably won't need to learn to continue contributing, but those who want it can have it.

There are also some minor JSDoc improvements scatter here which is great.

> In 97e6809, I got npm run typecheck to pass at the expense of IE11 support.

Does that mean that the browser build won't work on IE11, that would be a pity, but not a dealbreaker for me. While I haven't had a customer mention IE11 recently, up until only a year or so ago it was still expected sometimes.

SKalt:
Glad you like it!

> >  In 97e6809, I got npm run typecheck to pass at the expense of IE11 support
>
> Does that mean that the browser build won't work on IE11 [?]

Yep, I added some ES6 syntax, specifically [`class`, `Array.map/.reduce`, and an `Object.assign`](https://github.com/mapbox/mapbox-sdk-js/commit/97e68093eaa08241547e9caa65a60e2439004640#diff-0b158b79f0de615248a84dbaad56d7e6R12), which won't work out-of-the box in IE11.  Hoever, it would be feasible to add a step to the `rollup` compilation that would compile the ES6 to ES5 using `babel` or even `buble`.

PR #362, opened 2020-04-18
Correctly handle arrays containing zero values in appendQueryObject

sheridanvk:
Hi there - I ran into a bug when I tried to use a MapboxGeocoder object on a map initialised to a point in London (lng: 0, lat: 51.5). It turned out that after fixing that issue (draft PR here: https://github.com/mapbox/mapbox-gl-geocoder/pull/350) there was also a similar bug in the request that the geocoder triggered from this SDK. I believe this PR will solve it, but please let me know if I should do anything differently here!

Without the change I added, appendQueryObject returns `?proximity=51.5`, which breaks because proximity expects an input with two values, not one. With the change, it correctly returns `?proximity=0%2C51.5`.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -1448,7 +1448,7 @@ See the [public documentation][230].
   - `config.language` **[Array][206]&lt;[string][197]>?** Specify the language to use for response text and, for forward geocoding, query result weighting.
      Options are [IETF language tags][233] comprised of a mandatory
      [ISO 639-1 language code][234] and optionally one or more IETF subtags for country or script.
-  - `config.routing` **[boolean][198]?** Specify whether to request additional metadata about the recommended navigation destination. Only applicable for address features.
+  - `config.routing` **[boolean][198]** Specify whether to request additional metadata about the recommended navigation destination. Only applicable for address features. (optional, default `false`)
 
 #### Examples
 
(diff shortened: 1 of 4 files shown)


docs/services.md
```diff
@@ -1448,7 +1448,7 @@ See the [public documentation][230].
   - `config.language` **[Array][206]&lt;[string][197]>?** Specify the language to use for response text and, for forward geocoding, query result weighting.
      Options are [IETF language tags][233] comprised of a mandatory
      [ISO 639-1 language code][234] and optionally one or more IETF subtags for country or script.
-  - `config.routing` **[boolean][198]?** Specify whether to request additional metadata about the recommended navigation destination. Only applicable for address features.
+  - `config.routing` **[boolean][198]** Specify whether to request additional metadata about the recommended navigation destination. Only applicable for address features. (optional, default `false`)
```
sheridanvk, line 1451: I didn't make this change; I assume it came in through the pre-commit hooks. It seems to be related to the change that was made in the latest commit on master. Happy to just leave it in here or have you handle it separately and I can rebase my fork against that change; just let me know.

andrewharvey: That's fine, the docs are generated automatically and for various reasons (like the person before not running it) there can be unexpected changes. I wouldn't worry about it.

PR #363, opened 2020-04-22
set default value of getStyle metadata option

andrewharvey:


services/styles.js
```diff
@@ -342,7 +342,10 @@ Styles.getStyleSprite = function(config) {
     xtend(
       {
         method: 'GET',
-        path: '/styles/v1/:ownerId/:styleId' + (config.draft ? '/draft' : '') + fileName,
+        path:
+          '/styles/v1/:ownerId/:styleId' +
+          (config.draft ? '/draft' : '') +
+          fileName,
```
andrewharvey, line 348: looks like the commit hook made this change and not me

Issue #364, opened 2020-04-22
should we always add ?fresh=true to style requests


andrewharvey:
I've encountered instances where getStyle would return an outdated version of the style for a few minutes after updating and publishing changes in Studio. I've seen `?fresh=true` sometimes added as an argument to workaround this, but doesn't look to be a documented flag.

Given when working with getStyle you usually want the freshest style should we automatically pass this?

samanpwbb:
Hey @andrewharvey 👋. We intentionally do not make fresh=true default for two reasons: 

1. Requests that use this param bypass Mapbox's caching, and request times can therefore be quite a bit slower.
2.Requests that use this param have a significantly lower rate limit than cached requests.

There are a few valid use-cases for `fresh=true`, for example, to preview a change in your app if you're co-developing a style and an app at the same time, but it should never be used in production / high concurrency environments.

Hope this helps clarify things, and we'll get some documentation up around fresh=true soon.

okdistribute:
Hi, thanks for that @samanpwbb. I am in need of fresh=true for previewing changes in the app where we're using a style.

Is this possible to pass this directly to the getStyle function as part of the object?

PR #365, opened 2020-04-28
feat: add 'fresh=true' for getStyle function

okdistribute:
https://github.com/mapbox/mapbox-sdk-js/issues/364

I am in need of fresh=true for previewing changes in the app where we're using a style.

From @samanpwbb 

>    *  Requests that use this param bypass Mapbox's caching, and request times can therefore be quite a bit slower.
>    * Requests that use this param have a significantly lower rate limit than cached requests.
> 
> There are a few valid use-cases for fresh=true, for example, to preview a change in your app if you're co-developing a style and an app at the same time, but it should never be used in production / high concurrency environments.
> 
> Hope this helps clarify things, and we'll get some documentation up around fresh=true soon.

### DIFF
--- a/services/styles.js
+++ b/services/styles.js
@@ -23,7 +23,7 @@ var Styles = {};
  * @param {string} [config.ownerId]
  * @param {boolean} [config.metadata] - If true, `mapbox:` specific metadata will be preserved
  * @param {boolean} [config.draft=false] - If `true` will retrieve the draft style, otherwise will retrieve the published style.
- * @param {boolean} [config.fresh=false] - If `true` will bypass the style cache, with a lower rate limit than cached requests.
+ * @param {boolean} [config.fresh=false] - If `true`, will bypass the cached version of the style. Fresh style requests have a lower rate limit than cached requests and may have a higher latency. `fresh=true` should never be used in production or high concurrency environments.
  * @return {MapiRequest}
  *
  * @example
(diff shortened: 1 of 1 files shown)


andrewharvey:
I'd find this useful to for, if it's acceptable to use the fresh flag in a low concurrency setting (not delivering maps to end users), then we should support it.

Certainly if it's going to be documented and be part of the public API then supporting it here makes sense in my opinion.

services/styles.js
```diff
@@ -23,6 +23,7 @@ var Styles = {};
  * @param {string} [config.ownerId]
  * @param {boolean} [config.metadata] - If true, `mapbox:` specific metadata will be preserved
  * @param {boolean} [config.draft=false] - If `true` will retrieve the draft style, otherwise will retrieve the published style.
+ * @param {boolean} [config.fresh=false] - If `true` will bypass the style cache, with a lower rate limit than cached requests.
```
danswick, line 26: I think it would be really helpful to future users to include some version of @samanpwbb's disclaimer in the JSDoc. What do you think about something like this? 


```suggestion
 * @param {boolean} [config.fresh=false] - If `true`, will bypass the cached version of the style. Fresh style requests have a lower rate limit than cached requests and may have a higher latency. `fresh=true` should never be used in production or high concurrency environments.
```

okdistribute:
@danswick thanks! I accepted the requested change.

services/styles.js
andrewharvey: Just on this wording I think just saying not for use in high concurrency environments is enough, we could drop the "in production" wording. One of the common use cases of this SDK is to mutate styles within your own Mapbox account in which case you often need to get the current resource. eg. imagine if Mapbox Studio didn't use fresh=true, so for some production applications fresh=true is absolutely the right thing to do.

PR #373, opened 2020-05-26
task: update `uploads.createUpload` function params

### LINKED ISSUE #336: confused about `mapId` param in uploads
Is there a specific reason why `mapId` is required in `uploadsClient.createUpload` ? I noticed that the API itself seems to want a [`tileset` param](https://docs.mapbox.com/api/maps/#create-an-upload)

I am looking to replace a tileset with new vector tiles so currently I am specifying the tileset id as `mapId` ... is this correct?

Thank you! 🙏

theworkflow:
## Overview
Update `createUpload` parameters to match [updated documentation](https://docs.mapbox.com/api/maps/#create-an-upload). Supports backwards compatibility.

- `config.mapId` will overwrite `config.tileset` if set
- `config.tilesetName` will overwrite `config.name` if set

### Changed
- Add keys to config object passed into `createUpload`
  - tileset
  - name
  - private

- Update request body for `createUpload`
```js
// before
requestBody = {
  tileset: config.mapId,
  url: config.url,
  name: config.titlesetName
}

// after
requestBody = {
  tileset: config.titleset,
  url: config.url,
  name: config.name,
  private: config.private // defaults to true if not set
}
```

closes #336

### DIFF
--- a/services/uploads.js
+++ b/services/uploads.js
@@ -2,6 +2,7 @@
 
 var v = require('./service-helpers/validator');
 var createServiceFactory = require('./service-helpers/create-service-factory');
+var pick = require('./service-helpers/pick');
 
 /**
  * Uploads API service.
@@ -83,13 +84,11 @@ Uploads.createUploadCredentials = function() {
  * See the [corresponding HTTP service documentation](https://docs.mapbox.com/api/maps/#create-an-upload).
  *
  * @param {Object} config
- * @param {string} config.mapId - The map ID to create or replace in the format `username.nameoftileset`.
+ * @param {string} config.tileset - The tileset ID to create or replace, in the format `username.nameoftileset`.
  *   Limited to 32 characters (only `-` and `_` special characters allowed; limit does not include username).
- * @param {string} config.url - Either of the following:
- *   - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
- *   - The `mapbox://` URL of an existing dataset that you'd like to export to a tileset.
- *     This should be in the format `mapbox://datasets/{username}/{datasetId}`.
- * @param {string} [config.tilesetName] - Name for the tileset. Limited to 64 characters.
+ * @param {string} config.url - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
+ * @param {string} [config.name] - The name of the tileset. Limited to 64 characters.
+ * @param {boolean} [config.private=true] - A boolean that describes whether the tileset must be used with an access token from your Mapbox account. Default is true.
  * @return {MapiRequest}
  *
  * @example
@@ -103,8 +102,10 @@ Uploads.createUploadCredentials = function() {
  *   url: '{s3 url}'
  * };
  * uploadsClient.createUpload({
- *   mapId: `${myUsername}.${myTileset}`,
- *   url: credentials.url
+ *   titleset: `${myUsername}.${myTileset}`,
+ *   url: credentials.url,
+ *   name: 'my uploads name',
+ *   private: true
  * })
  *   .send()
  *   .then(response => {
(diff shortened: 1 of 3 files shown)


andrewharvey:
Thanks for the PR. Looks like it addresses https://docs.mapbox.com/api/changelog/#november-5-2019.

I linked this to #336 and agree with updating away from mapId which isn't used much anymore. Unfortuantly there is little consistancy in the API for example https://docs.mapbox.com/api/maps/#delete-tileset uses `tileset_id` and so this SDK uses `tilesetId` at https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/services.md#deletetileset.

It would be nice if these were all consistent, but without the actual API being consistent, I'm okay with either option (following upstream API `tileset` or the rest of the SDK `tilesetId`.

However I think that should be done as a backwards compatible change so we don't need to make a breaking change. Basically if `tileset` isn't passed then also check `mapId`.

services/uploads.js
```diff
@@ -83,13 +83,11 @@ Uploads.createUploadCredentials = function() {
...
- * @param {string} config.mapId - The map ID to create or replace in the format `username.nameoftileset`.
+ * @param {string} config.tileset - The tileset ID to create or replace, in the format `username.nameoftileset`.
  *   Limited to 32 characters (only `-` and `_` special characters allowed; limit does not include username).
- * @param {string} config.url - Either of the following:
- *   - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
- *   - The `mapbox://` URL of an existing dataset that you'd like to export to a tileset.
- *     This should be in the format `mapbox://datasets/{username}/{datasetId}`.
- * @param {string} [config.tilesetName] - Name for the tileset. Limited to 64 characters.
+ * @param {string} config.url - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
+ * @param {string} config.name - Optional. The name of the tileset. Limited to 64 characters.
```
andrewharvey, line 89:
```suggestion
 * @param {string} [config.name] - The name of the tileset. Limited to 64 characters.
```

use JSDoc to mark it as optional and then the generated docs will add optional automatically

services/uploads.js
```diff
@@ -83,13 +83,11 @@ Uploads.createUploadCredentials = function() {
...
+ * @param {boolean} config.private - Optional. A boolean that describes whether the tileset must be used with an access token from your Mapbox account. Default is true.
```
andrewharvey, line 90:
```suggestion
 * @param {boolean} [config.private=true] - A boolean that describes whether the tileset must be used with an access token from your Mapbox account.
```

Using JSDoc to set optional and default

services/uploads.js
```diff
@@ -113,19 +113,29 @@ Uploads.createUploadCredentials = function() {
  */
 Uploads.createUpload = function(config) {
   v.assertShape({
-    mapId: v.required(v.string),
+    tileset: v.required(v.string),
     url: v.required(v.string),
-    tilesetName: v.string
```
andrewharvey, line 118: if this can also be handled in a backwards compatible way to prevent a breaking change

services/uploads.js
```diff
@@ -113,19 +113,29 @@ Uploads.createUploadCredentials = function() {
...
+
   return this.client.createRequest({
     method: 'POST',
     path: '/uploads/v1/:ownerId',
-    body: {
-      tileset: config.mapId,
-      url: config.url,
-      name: config.tilesetName
-    }
+    body: body
```
andrewharvey, line 138: To keep the codebase consistent could you use pick to set the body? See https://github.com/mapbox/mapbox-sdk-js/blob/fc74a74f6c48d7373aec06548379f0c534f4d298/services/tilesets.js#L40-L58 for an example of how it's done throughout the rest of the codebase.

services/uploads.js
```diff
@@ -113,19 +113,29 @@ Uploads.createUploadCredentials = function() {
...
+    tileset: config.tileset,
+    url: config.url
+  };
+
+  if (config.name) body.name = config.name;
+
+  if (config.private === false) {
+    body.private = false;
+  } else {
+    body.private = true;
```
andrewharvey, line 132: See https://github.com/mapbox/mapbox-sdk-js/blob/fc74a74f6c48d7373aec06548379f0c534f4d298/services/tilesets.js#L262-L277 for an example of a boolean option. You should just be able to pick out private without needing this.

theworkflow: The `private` key defaults to `true`. The `pick` function won't default the boolean value to true. I could write my own `filter` function, but I think it would be more verbose instead of the logic that I have now. Unless I'm hardcoding names in the `filter` function, I would have to do some type of type checking to default `private` to `true`. Thoughts on this?

services/uploads.js
theworkflow: [This comment](https://github.com/mapbox/mapbox-sdk-js/pull/373#discussion_r430472960) speaks on using `pick` a bit. I think more work would be required to use `pick`.

services/uploads.js
```diff
@@ -83,13 +84,16 @@ Uploads.createUploadCredentials = function() {
...
  *   Limited to 32 characters (only `-` and `_` special characters allowed; limit does not include username).
- * @param {string} config.url - Either of the following:
- *   - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
- *   - The `mapbox://` URL of an existing dataset that you'd like to export to a tileset.
- *     This should be in the format `mapbox://datasets/{username}/{datasetId}`.
- * @param {string} [config.tilesetName] - Name for the tileset. Limited to 64 characters.
+ * @param {string} config.url - HTTPS URL of the S3 object provided by [`createUploadCredentials`](#createuploadcredentials)
+ * @param {string} [config.name] - The name of the tileset. Limited to 64 characters.
+ * @param {boolean} [config.private=true] - A boolean that describes whether the tileset must be used with an access token from your Mapbox account. Default is true.
+ * @param {string} [config.mapId] - The map ID to create or replace, in the format `username.nameoftileset`.
```
andrewharvey, line 92: If this is just here for backwards compatibility, I'd not include it in the current JSDoc since it will just lead to confusion on what to use. Same for tilesetName.

theworkflow:
Sorry for the extra ping. Still looking to get this into the next release @andrewharvey

PR #384, opened 2020-06-28
try and build bundle as part of automated test suit to catch any build issues

### LINKED ISSUE #381: add a test for bundle building
breaking out from #380, there was an issue building the bundle which wasn't caught by the automated testing, we should include the bundle build in the automated testing to try and catch any issues earlier.

andrewharvey:
closes #381


andrewharvey:
This test fails, but as expected since master is broken.

PR #397, opened 2020-09-16
Allow the client to work inside a web worker

MarnixBouhuis:
Hi,

Currently this sdk doesn't work when you call methods from inside a webworker.
This is because a worker doesn't have access to the window object. In a WorkerGlobalScope you have to use "self" instead of window. 

This only requires 1-2 lines of code to be changed.

Kind regards,
Marnix


MarnixBouhuis:
Last build also looks like it stalled. Looks like the same issue as described here: https://github.com/travis-ci/travis-ci/issues/9772

PR #399, opened 2020-09-23
Update services.md

kasi1999999:
Typo fix


andrewharvey:
Thanks for the PR, however since services.md is automatically updated, it needs to be updated in the source at https://github.com/mapbox/mapbox-sdk-js/blob/bb84221720f13404a9ae883849b1be1c3e06042f/services/map-matching.js#L114

andrewharvey:
this was updated in #400 so closing this

Issue #392, opened 2020-07-30
Upload error


jlumia:
I am getting this error when calling the createUpload method. I didn't know if this was the right place to put this comment.

`      await uploadsClient
        .createUpload({
          mapId: `${userName}.${sources[0].tilesetName}`,
          url: s3Creds.url,
          tilesetName: sources[0].fileName
        })
        .send();`

`MapiError {
  message: 'The request body contains the invalid property "private".',
  type: 'HttpError',
  statusCode: 422,
  request:
   MapiRequest {
     id: 34,
     _options:
      { method: 'POST',
        path: '/uploads/v1/:ownerId',
        body:
         { url:
            'https://tilestream-tilesets-production.s3.amazonaws.com/<myamazonurl>',
           tileset: '<mytileset>',
           name: '<mysource>',
           private: true } },
     emitter: EventEmitter { _events: {}, _eventsCount: 0 },
     client:
      NodeClient {
        accessToken:
         '<mytoken>',
        origin: 'https://api.mapbox.com' },
     response: null,
     error: [Circular],
     sent: true,
     aborted: false,
     path: '/uploads/v1/:ownerId',
     method: 'POST',
     origin: 'https://api.mapbox.com',
     query: {},
     params: {},
     body:
      { url:
         'https://tilestream-tilesets-production.s3.amazonaws.com/<myamazonurl>',
        tileset: '<mytileset>',
        name: '<mysource>',
        private: true },
     file: null,
     encoding: 'utf8',
     sendFileAs: null,
     headers: { 'content-type': 'application/json' } },
  body:
   { message: 'The request body contains the invalid property "private".' } 
}`

andrewharvey:
> "mapbox": "^1.0.0-beta9"

That's a very old version, could you try upgrading to the latest `@mapbox/mapbox-sdk` package instead of the old `mapbox` package?

jlumia:
@andrewharvey Sorry, that was some old cruft from the package json. I am actually using `"@mapbox/mapbox-sdk": "^0.11.0"` and leveraging it in my .js file `const mapboxUploads = require('@mapbox/mapbox-sdk/services/uploads');`

andrewharvey:
I can replicate this with

```js
const MapboxUploads = require('@mapbox/mapbox-sdk/services/uploads');
const uploadsService = MapboxUploads({ accessToken: 'sk.' });

uploadsService.createUpload({
    mapId: 'username.tileset',
    url: 'test',
[...]

PR #401, opened 2020-10-12
Fix typo in docs

halafi:


docs/services.md
```diff
@@ -658,7 +658,7 @@ const credentials = {
   url: '{s3 url}'
 };
 uploadsClient.createUpload({
-  titleset: `${myUsername}.${myTileset}`,
+  tileset: `${myUsername}.${myTileset}`,
   url: credentials.url,
   name: 'my uploads name',
   private: true
```
halafi, line 664: the `private` property throws another error:

`message: 'The request body contains the invalid property "private".'`

andrewharvey: That's probably #392

docs/services.md
```diff
@@ -658,7 +658,7 @@ const credentials = {
   url: '{s3 url}'
 };
 uploadsClient.createUpload({
-  titleset: `${myUsername}.${myTileset}`,
+  tileset: `${myUsername}.${myTileset}`,
```
andrewharvey, line 661: Thanks for the PR, the md files are autogenerated from the JS files, so the corresponding line is at https://github.com/mapbox/mapbox-sdk-js/blob/main/services/uploads.js#L105, could you update it there too?

docs/services.md
halafi: @andrewharvey yep, I think there is a way to satisfy the spec by not setting default value for private on the client, anyway left a comment in #392

docs/services.md
halafi: done

andrewharvey (approved):
Thanks

PR #404, opened 2020-11-18
replaceTilesetSource method

mapsam:
This adds the `Tilesets.replaceTilesetSource` method to the Tilesets service. https://docs.mapbox.com/api/maps/#replace-a-tileset-source

I'm reading through the development requirements, and think I need to verify a few things:

* https://github.com/mapbox/mapbox-sdk-js/blob/main/docs/development.md#service-method-naming-conventions suggests PUT methods should be named `put*`, so should I rename this to `Tilesets.putTilesetSource`? Happy to change it!
* I considered adding a `replace: true` option to `Tilesets.createTilesetSource` method instead of an entirely new method. This would effectively change POST to PUT, but I wasn't able to find any instances of this pattern in the rest of the SDK. Thoughts on this idea?

Also massive h/t to @andrewharvey for adding the Tilesets service to the SDK originally. So good! 

cc @katydecorah @andrewharvey

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -38,6 +38,7 @@
   - [deleteTileset](#deletetileset)
   - [tileJSONMetadata](#tilejsonmetadata)
   - [createTilesetSource](#createtilesetsource)
+  - [replaceTilesetSource](#replacetilesetsource)
   - [getTilesetSource](#gettilesetsource)
   - [listTilesetSources](#listtilesetsources)
   - [deleteTilesetSource](#deletetilesetsource)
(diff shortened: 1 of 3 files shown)


andrewharvey (dismissed):
This is great thanks, looks like just needs a rebase against main to rebuild the docs.

jboolean:
This would be a valuable addition for me. Can we get this in?

andrewharvey:
rebased against main with rebuilt docs

andrewharvey:
> https://github.com/mapbox/mapbox-sdk-js/blob/main/docs/development.md#service-method-naming-conventions suggests PUT methods should be named put*, so should I rename this to Tilesets.putTilesetSource? Happy to change it!

I think best to leave it as `replaceTilesetSource` to be consistent with the API, despite going against the service naming conventions.

> I considered adding a replace: true option to Tilesets.createTilesetSource method instead of an entirely new method. This would effectively change POST to PUT, but I wasn't able to find any instances of this pattern in the rest of the SDK. Thoughts on this idea?

Definitely a separate method sounds best to me, as it mirrors the main API and the method name is immediately clear on what it does.

luskin:
Can we please get this merged in.

PR #405, opened 2020-12-08
Removing "private" from "createUpload"

boxman0617:
Based on https://github.com/mapbox/mapbox-sdk-js/issues/392

```
MapiError {
  message: 'The request body contains the invalid property "private".',
  type: 'HttpError',
  statusCode: 422,
  request: MapiRequest {
    id: 1,
    _options: { method: 'POST', path: '/uploads/v1/:ownerId', body: [Object] },
    emitter: EventEmitter {
      _events: Events <Complex prototype> {},
      _eventsCount: 0
    },
    client: NodeClient {
      accessToken: '{accessToken}',
      origin: 'https://api.mapbox.com'
    },
    response: null,
    error: [Circular],
    sent: true,
    aborted: false,
    path: '/uploads/v1/:ownerId',
    method: 'POST',
    origin: 'https://api.mapbox.com',
    query: {},
    params: {},
    body: {
      tileset: '{username}.{name},
      url: '{url}',
      name: 'Random Name,
      private: true
    },
    file: null,
    encoding: 'utf8',
    sendFileAs: null,
    headers: { 'content-type': 'application/json' }
  },
  body: {
    message: 'The request body contains the invalid property "private".'
  }
}
```

I removed all of the instances of the use "private" and updated the tests.

Tested the request without private and the API was successful.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -643,7 +643,6 @@ See the [corresponding HTTP service documentation][218].
       Limited to 32 characters (only `-` and `_` special characters allowed; limit does not include username).
   - `config.url` **[string][201]** HTTPS URL of the S3 object provided by [`createUploadCredentials`][39]
   - `config.name` **[string][201]?** The name of the tileset. Limited to 64 characters.
-  - `config.private` **[boolean][202]** A boolean that describes whether the tileset must be used with an access token from your Mapbox account. Default is true. (optional, default `true`)
 
 #### Examples
 
@@ -661,7 +660,6 @@ uploadsClient.createUpload({
   titleset: `${myUsername}.${myTileset}`,
   url: credentials.url,
   name: 'my uploads name',
-  private: true
 })
   .send()
   .then(response => {
--- a/services/__tests__/uploads.test.js
+++ b/services/__tests__/uploads.test.js
@@ -35,8 +35,7 @@ describe('createUpload', () => {
     uploads.createUpload({
       tileset: 'username.nameoftileset',
       url: 'http://{bucket}.s3.amazonaws.com/{key}',
-      name: 'dusty_devote',
-      private: false
+      name: 'dusty_devote'
     });
 
     expect(tu.requestConfig(uploads)).toEqual({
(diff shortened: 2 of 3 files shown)


andrewharvey:
Thanks for the PR. According to https://docs.mapbox.com/api/maps/uploads/#create-an-upload `private` is still an allowed parameter. I'd like some :eyes: from the API team to look at this first to comment on if it's an API issue, or something the API changed, therefore the docs need changing, then something we can fix here.

I'm hesitant to roll with this fix without knowing what's going on on the API side.

boxman0617:
@andrewharvey Totally agree. I'd like to know what changed. The docs https://docs.mapbox.com/api/maps/#create-an-upload state that private is an optional field, but any time that is added to the request the `message: 'The request body contains the invalid property "private".',` message is returned from the API.

I tried it on my project without the "private" field and it worked just fine.

Let me know what the end result of this is as I would either like to merge this PR and update my deps or keep the workaround i have till a fix is made. 

Thank you!

andrewharvey:
@katydecorah @danswick Do you know if this is a problem with the API docs, or the API? Trying to workout if this is something the SDK should fix or the API/docs?

boxman0617:
@andrewharvey @katydecorah @danswick Any word on this?

paulJordaan:
@andrewharvey just ran into the same issue that this PR addresses. Can confirm `createUpload` service fails because of the default `private` being passed. Any update would appreciated.

andrewharvey:
> @andrewharvey just ran into the same issue that this PR addresses. Can confirm createUpload service fails because of the default private field being passed in the body. Any update would appreciated.

Because of the conflict between what the API documentation says (which says you can pass `private`)  and the actual behaviour of the API (which returns an error when you do). I'm not that keen on making a decision on this, because it's not clear if this is just a bug in the API that should be fixed, or it was a breaking change made to the API which we should address here.

As I said earlier on this thread, personally I'm hesitant to make a change without a resolution from Mapbox about this issue. To progress this, I'd suggest to raise a ticket with Mapbox through the official support channel, and if you do get an update, please share back here.

millzpaugh (approved):
Thanks @boxman0617 for updating our SDK - I've confirmed that the `private` option is not in fact accepted in the create upload endpoint for the Uploads API. We'll update our docs. cc @kbauhausness

PR #408, opened 2020-12-31
Fix typo in `createUpload` example

paulJordaan:
Correct example to use `tileset` instead on `titleset`


andrewharvey:
Thanks. The .md file is auto generated from the source code, if you could update this in the js file please then I can merge.

paulJordaan:
Hi @andrewharvey, I've update the js file as requested

andrewharvey (approved):

PR #411, opened 2021-01-27
adding bbox and padding

shabnomnom:
Adding [BBox and Padding](https://docs.mapbox.com/api/maps/static-images/#retrieve-a-static-map-from-a-style) into the SDK. Let me know if you have any feedback!

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -417,12 +417,16 @@ SDK returned.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
   - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the
-      bounds of the overlay(s). Otherwise, the maps' position is described by an object
-      with the following properties:
-      `coordinates` (required): [`coordinates`][191] for the center of image.
-      `zoom` (required): Between 0 and 20.
+     bounds of the overlay(s). Another option for config.position is a bounding box object. 
+     `bbox` (optional): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+     corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
+     Otherwise the maps' position is described by an object with the following properties:
+      `coordinates` (optional): [`coordinates`][191] for the center of image.
+      `zoom` (optional): Between 0 and 20.
       `bearing` (optional): Between 0 and 360.
       `pitch` (optional): Between 0 and 60.
+  - `config.padding` **[string][201]** A string value that denotes the minimum padding per side of the image. 
+      This can only be used with auto or bbox. The value resembles the CSS specification for padding and accepts 1-4 integers without units
   - `config.overlays` **[Array][210]&lt;Overlay>?** Overlays should be in z-index
       order: the first in the array will be on the bottom; the last will be on
       the top. Overlays are objects that match one of the following types:
@@ -453,6 +457,24 @@ staticClient.getStaticImage({
     coordinates: [12, 13],
     zoom: 4
   }
+})
+  .send()
+  .then(response => {
+    const image = response.body;
+  });
+```
+
+```javascript
+staticClient.getStaticImage({
+  ownerId: 'mapbox',
+  styleId: 'streets-v11',
+  width: 200,
+  height: 300,
+  position: {
+    // position as a bounding box
+    bbox: [-77.04,38.8,-77.02,38.91],
+  }, 
+ padding: '4'
 })
   .send()
   .then(response => {
--- a/docs/services.md
+++ b/docs/services.md
@@ -417,16 +417,16 @@ SDK returned.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
   - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
-     If an object, it can contain either a bounding box or a coordinate and a zoom as the required parameters.  
-     `bbox` (required): An array of coordinate pairs, with the first coordinate pair referring to the southwestern
+     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
+     `bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
      corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
-     If you do not use `bbox`, the map's position is described by an object with the following properties:
+     Otherwise the maps' position is described by an object with the following properties:
       `coordinates` (required): [`coordinates`][191] for the center of image.
       `zoom` (required): Between 0 and 20.
       `bearing` (optional): Between 0 and 360.
       `pitch` (optional): Between 0 and 60.
   - `config.padding` **[string][201]** A string value that denotes the minimum padding per side of the image. 
-      This can only be used with `auto` or `bbox`. The value resembles the CSS specification for padding and accepts 1-4 integers without units
+      This can only be used with auto or bbox. The value resembles the CSS specification for padding and accepts 1-4 integers without units
   - `config.overlays` **[Array][210]&lt;Overlay>?** Overlays should be in z-index
       order: the first in the array will be on the bottom; the last will be on
       the top. Overlays are objects that match one of the following types:
--- a/docs/services.md
+++ b/docs/services.md
@@ -417,16 +417,16 @@ SDK returned.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
   - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
-     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
-     `bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+     If an object, it can contain either a bounding box or a coordinate and a zoom as the required parameters.  
+     `bbox` (required): An array of coordinate pairs, with the first coordinate pair referring to the southwestern
      corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
-     Otherwise the maps' position is described by an object with the following properties:
+     If you do not use `bbox`, the map's position is described by an object with the following properties:
       `coordinates` (required): [`coordinates`][191] for the center of image.
       `zoom` (required): Between 0 and 20.
       `bearing` (optional): Between 0 and 360.
       `pitch` (optional): Between 0 and 60.
   - `config.padding` **[string][201]** A string value that denotes the minimum padding per side of the image. 
-      This can only be used with auto or bbox. The value resembles the CSS specification for padding and accepts 1-4 integers without units
+      This can only be used with `auto` or `bbox`. The value resembles the CSS specification for padding and accepts 1-4 integers without units
   - `config.overlays` **[Array][210]&lt;Overlay>?** Overlays should be in z-index
       order: the first in the array will be on the bottom; the last will be on
       the top. Overlays are objects that match one of the following types:
--- a/docs/services.md
+++ b/docs/services.md
@@ -416,13 +416,13 @@ SDK returned.
   - `config.styleId` **[string][201]** The map's style ID.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
-  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the
-     bounds of the overlay(s). Another option for config.position is a bounding box object. 
-     `bbox` (optional): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
+     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
+     `bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
      corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
      Otherwise the maps' position is described by an object with the following properties:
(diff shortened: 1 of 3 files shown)


services/static.js
```diff
@@ -179,13 +202,15 @@ Static.getStaticImage = function(config) {
       v.oneOfType(
         v.oneOf('auto'),
         v.strictShape({
-          coordinates: v.required(v.coordinates),
```
andrewharvey, line 182: I think we could improve this, right now it will let you pass an empty object for position.

Instead could we do oneOfType:
- coordinates, zoom required, rest optional
- bbox require, rest optional

andrewharvey:
Thanks, just yesterday I wasted time implementing geo-viewport and adding in a padding to the library, because I was relying on the outdated SDK docs here instead of realising the Static API actually already supported a bbox and padding. So glad I saw this now.

shabnomnom:
👋🏽 The padding and the bbox are fairly new features in the Static API too. They both were added on [December 10th, 2020.](https://docs.mapbox.com/api/maps/changelog/#static-images-api)

services/static.js
shabnomnom: thanks for your review, let me do some testing and get back to you!

shabnomnom: @andrewharvey, Can you explain what do you mean by "it will let you pass an empty object?" When I test this on my side with `position: {}`, I receive the following error: 
```
../mapbox-sdk-js/services/static.js:323
    .concat([
     ^
TypeError: Cannot read property 'concat' of undefined
```
Even though this is not a specific e error message but it does not pass beyond the `encodePosition()` function. Moving forward I can add an error to catch this scenario and return a more helpful message inside the encodePosition()` function. 
Another option here is to configure the position as you suggested:    
```
position: v.required(
v.oneOfType(
        v.oneOf('auto'),
        v.strictShape({
          coordinates: v.required(v.coordinates)
          zoom: v.required(v.range([0, 20])),
          bearing: v.range([0, 360]),
          pitch: v.range([0, 60]),
          bbox: v.arrayOf(v.number)
        }),
        v.strictShape({bbox: v.required(v.arrayOf(v.number))})
      )
```
With this method, we are relying on the[ fusspot library](https://github.com/mapbox/fusspot#vnumber) to validate the position object. In `position:{}` scenario , fusspot will return an error `Error: position.bbox is required` which is not necessarily true because only `positon` is required and it could be in a form of `bbox`,`auto` or `coordinates & zoom`. In case of an additional location parameter with the bbox position such as: `position: { bbox:[-77.04,38.89,-77.02,38.91], pitch: 4}` , the fusspot will simply return a an error related to the first `oneOfType()` arguments. In this case: `Error: position must be "auto"` which is not necessarily helpful either. 

Let me know if this is helpful or if you have any other suggestions!

shabnomnom: Another thing to consider is that it is not possible to use the `bbox` with any other location parameters such as `zoom`, `pitch`, and `bearing`.

services/static.js
```diff
@@ -269,7 +316,9 @@ Static.getStaticImage = function(config) {
 
 function encodePosition(position) {
   if (position === 'auto') return 'auto';
-
+  if (position.bbox instanceof Array) {
+    return '[' + String(position.bbox) + ']';
+  }
```
jaericson, line 321: Would JSON.stringify work here?

shabnomnom: yes, it does. I just implemented it.

services/__tests__/static.test.js
```diff
@@ -39,6 +39,100 @@ describe('getStaticImage', () => {
...
+      method: 'GET',
+      path:
+        '/styles/v1/:ownerId/:styleId/static/[-77.04,38.8,-77.02,38.91]/200x300',
+      query: {},
+      params: { ownerId: 'mapbox', styleId: 'streets-v10' },
+      encoding: 'binary'
+    });
+  });
+
+  test('padding', () => {
```
riastrad, line 62: [request] this is great, can we also add tests for other forms of padding? (i.e. 1, 3, & 4 comma delimited values)

services/__tests__/static.test.js
```diff
@@ -39,6 +39,100 @@ describe('getStaticImage', () => {
...
+    expect(tu.requestConfig(service)).toEqual({
+      method: 'GET',
+      path: '/styles/v1/:ownerId/:styleId/static/auto/200x300',
+      query: { padding: '3,4' },
+      params: { ownerId: 'mapbox', styleId: 'streets-v10' },
+      encoding: 'binary'
+    });
+  });
+
+  test('bbox with pitch and bearing', () => {
```
riastrad, line 101: [question] padding should be compatible with bbox, yes? let's add a test for that as well.

services/static.js
```diff
@@ -182,10 +205,13 @@ Static.getStaticImage = function(config) {
           coordinates: v.required(v.coordinates),
           zoom: v.required(v.range([0, 20])),
           bearing: v.range([0, 360]),
-          pitch: v.range([0, 60])
-        })
+          pitch: v.range([0, 60]),
+          bbox: v.arrayOf(v.number)
```
riastrad, line 209: [blocker] I don't think `bbox:` should be included as a value here.

services/static.js
```diff
@@ -269,7 +316,9 @@ Static.getStaticImage = function(config) {
 
 function encodePosition(position) {
   if (position === 'auto') return 'auto';
-
+  if (position.bbox instanceof Array) {
+    return JSON.stringify(position.bbox);
+  }
```
riastrad, line 321: [nit] this is a great, but technically redundant given the above assertion you have on L211 that bbox must always be an array of numbers. I think we can simplify this to 

```js
if (position.bbox) {
  return JSON.stringify(position.bbox)
}
```

And then add some error handling above to catch possible problems with the bbox (e.g. too many coordinates).

services/static.js
```diff
@@ -258,6 +287,24 @@ Static.getStaticImage = function(config) {
     );
   }
 
+  if (
+    config.padding !== undefined &&
+    (config.position !== 'auto' && config.position.bbox === undefined)
+  ) {
+    throw new Error(
+      'Padding can only be used with auto or with a bounding box as the position.'
```
riastrad, line 295:
```suggestion
      'Padding can only be used with auto or bbox as the position.'
```

services/static.js
```diff
@@ -258,6 +287,24 @@ Static.getStaticImage = function(config) {
     );
   }
 
+  if (
+    config.padding !== undefined &&
+    (config.position !== 'auto' && config.position.bbox === undefined)
```
riastrad, line 292: [nit] these parentheses are technically unnecessary

services/static.js
```diff
@@ -258,6 +287,24 @@ Static.getStaticImage = function(config) {
...
+  if (
+    config.position.bbox !== undefined &&
+    (config.position.pitch !== undefined ||
+      config.position.bearing !== undefined ||
+      config.position.zoom !== undefined)
+  ) {
+    throw new Error(
+      'The bounding box parameter cannot be used with additional location parameters, bearing, or pitch'
+    );
+  }
```
riastrad, line 307: [nit] If we properly assert on the shape of the `position` argument above, then this check should be mostly unnecessary.

services/static.js
```diff
@@ -25,13 +25,18 @@ var Static = {};
...
+ * @param {'auto'|Object} config.position - If `"auto"`, the viewport will fit the bounds of the overlay(s).
+ *  If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
+ *  ` bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+ *  corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
+ *  Otherwise the maps' position is described by an object with the following properties:
  *   `coordinates` (required): [`coordinates`](#coordinates) for the center of image.
  *   `zoom` (required): Between 0 and 20.
  *   `bearing` (optional): Between 0 and 360.
  *   `pitch` (optional): Between 0 and 60.
+ *
```
riastrad, line 37: [nit] can remove this unnecessary extra line

services/static.js
```diff
@@ -237,6 +263,9 @@ Static.getStaticImage = function(config) {
   if (config.layer_id !== undefined) {
     query.layer_id = config.layer_id;
   }
+  if (config.padding !== undefined) {
```
riastrad, line 266: [question] how do you feel about adding an assertion on the values of `padding` ahead of this line? Preferably some minimal guardrails to prevent users from generating URLs that are are very obviously wrong (e.g. `padding: 'a random string of letters'`)

riastrad (commented):
This is looking good so far! There are few things I thing we should address before rolling these updates out.

services/__tests__/static.test.js
shabnomnom: yes! I added the tests!

services/static.js
shabnomnom: Sounds good.

services/static.js
shabnomnom: 👋🏽  This seems to be handled pretty well on the API side. For example: 
https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/[-122,34,-121,45]/600x600?&padding=dgdf&access_token=TOKEN will have the error message of ` "Padding must be 1-4 numbers, comma-delimited."` Do you think it is still required to be handled on the SDK side too?

docs/services.md
```diff
@@ -416,13 +416,17 @@ SDK returned.
   - `config.styleId` **[string][201]** The map's style ID.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
-  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the
-      bounds of the overlay(s). Otherwise, the maps' position is described by an object
-      with the following properties:
+  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
+     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
+     `bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+     corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
```
HeyStenson, line 422:
```suggestion
     `bbox` (required): An array of coordinate pairs, with the first coordinate pair referring to the southwestern
     corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
```

docs/services.md
```diff
@@ -416,13 +416,17 @@ SDK returned.
   - `config.styleId` **[string][201]** The map's style ID.
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
-  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the
-      bounds of the overlay(s). Otherwise, the maps' position is described by an object
-      with the following properties:
+  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
+     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.
```
HeyStenson, line 420:
```suggestion
     If an object, it can contain either a bounding box or a coordinate and a zoom as the required parameters.  
```

docs/services.md
```diff
@@ -416,13 +416,17 @@ SDK returned.
...
   - `config.width` **[number][205]** Width of the image in pixels, between 1 and 1280.
   - `config.height` **[number][205]** Height of the image in pixels, between 1 and 1280.
-  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the
-      bounds of the overlay(s). Otherwise, the maps' position is described by an object
-      with the following properties:
+  - `config.position` **(`"auto"` \| [Object][200])** If `"auto"`, the viewport will fit the bounds of the overlay(s).
+     If an object, it could be either a bbox or a coordinate and a zoom as the required parameters.  
+     `bbox` (required): Is an array of coordinate pairs, with the first coordinate pair referring to the southwestern
+     corner of the box (the minimum longitude and latitude) and the second referring to the northeastern corner of the box (the maximum longitude and latitude).
+     Otherwise the maps' position is described by an object with the following properties:
```
HeyStenson, line 423:
```suggestion
     If you do not use `bbox`, the map's position is described by an object with the following properties:
```

docs/services.md
```diff
@@ -416,13 +416,17 @@ SDK returned.
...
       `coordinates` (required): [`coordinates`][191] for the center of image.
       `zoom` (required): Between 0 and 20.
       `bearing` (optional): Between 0 and 360.
       `pitch` (optional): Between 0 and 60.
+  - `config.padding` **[string][201]** A string value that denotes the minimum padding per side of the image. 
+      This can only be used with auto or bbox. The value resembles the CSS specification for padding and accepts 1-4 integers without units
```
HeyStenson, line 429:
```suggestion
      This can only be used with `auto` or `bbox`. The value resembles the CSS specification for padding and accepts 1-4 integers without units
```

HeyStenson (approved):
@shabnomnom Lookin' good! I added a few suggestions on the documentation end of things. 🍊

services/__tests__/static.test.js
```diff
@@ -39,6 +39,177 @@ describe('getStaticImage', () => {
...
+      method: 'GET',
+      path:
+        '/styles/v1/:ownerId/:styleId/static/[-77.04,38.8,-77.02,38.91]/200x300',
+      query: { padding: '1,2,3' },
+      params: { ownerId: 'mapbox', styleId: 'streets-v10' },
+      encoding: 'binary'
+    });
+  });
+
+  test('padding 1,2,3,4 pm for top, right, bottom and the left side  ', () => {
```
riastrad, line 139:
```suggestion
  test('padding 1,2,3,4 px for top, right, bottom and the left side  ', () => {
```

services/static.js
```diff
@@ -53,21 +57,39 @@ var Static = {};
  *   on the map image.
  * @return {MapiRequest}
  *
+ *
```
riastrad, line 60: [nit] this line is unncecessary

services/static.js
```diff
@@ -53,21 +57,39 @@ var Static = {};
...
- *   position: {
- *     coordinates: [12, 13],
- *     zoom: 4
- *   }
- * })
- *   .send()
- *   .then(response => {
- *     const image = response.body;
- *   });
+*   ownerId: 'mapbox',
```
riastrad, line 63: [nit] the spacing of all the `*` might is off by one for the next 29 lines

services/static.js
riastrad: 👍 summarizing our chat from earlier: the existing design of the SDK seems content to not validate any parts of the options provided aside from the standard type checking coming from fusspot and let the APIs' errors speak for themselves. That sounds good to me, so we should feel comfortable leaving this without more rigorous arg checks.

riastrad (approved):
some final nits, but this is looking great overall 🌮

PR #413, opened 2021-02-02
Fixes #412

### LINKED ISSUE #412: Argument mutation and incorrect result when attempting to reuse PathOverlay coordinates in getStaticImage config
When reusing a coordinates array with multiple path overlays in a getStaticImage config (for example to get an outlined-line effect), the second (and each subsequent alternating) path is not drawn in the correct position. It is drawn with the longitude and latitude for each point reversed — lat instead of long and long instead of lat. Also, the ```coordinates``` arrays are inadvertently mutated.

I've tracked down the source of this bug and will submit a pull request shortly.

```JavaScript
const coordinates = [
  [-97.686118, 30.395509],
  [-97.68584, 30.395919],
  [-97.686015, 30.396182],
  [-97.68591, 30.396206],
  [-97.685815, 30.396129]
];

const staticImageConfig = {
  ownerId,
  styleId,
  width: 375,
  height: 375,
  position: 'auto',
  highRes: true,
  overlays: [
    {
      path: {
        coordinates,
        strokeWidth: 6,
        strokeColor: '#fff'
      }
    },
    {
      path: {
        coordinates,
        strokeWidth: 3,
        strokeColor: '#157efb'
      }
    }
  ]
};

const request = staticClient.getStaticImage(staticImageConfig);
// coordinates has been mutated and each point has had its lat and long swapped
```

jjavery:
Array.prototype.reverse() reverses an array _in place_. This results in the unintended mutation of the coordinates property of encodePathOverlay()'s o (overlay) param. See a side-effect of this mutation in #412. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse

Instead, create a new array and initialize it with the reversed lat and long.

### DIFF
--- a/services/static.js
+++ b/services/static.js
@@ -431,7 +431,7 @@ function encodePathOverlay(o) {
   }
   // polyline expects each coordinate to be in reversed order: [lat, lng]
   var reversedCoordinates = o.coordinates.map(function(c) {
-    return c.reverse();
+    return [c[1], c[0]];
   });
   var encodedPolyline = polyline.encode(reversedCoordinates);
   result += '(' + encodeURIComponent(encodedPolyline) + ')';


andrewharvey (approved):
Thanks!

PR #415, opened 2021-04-01
updated package got to 11.8.2 as previous version require electron at…

tombombadilom:
… build and new version has electron removed

node -v
v12.20.1
"gatsby": "^3.0.1",
"react": "^17.0.1",
"react-map-gl": "^6.1.11",
"react-map-gl-geocoder": "^2.1.6",

here is what happend when you build @mapbox/mapbox-sdk with the old got version
```
ERROR #98124  WEBPACK
Generating SSR bundle failed
Can't resolve 'electron' in '{...}/node_modules/got'
If you're trying to use a package make sure that 'electron' is installed. If you're trying to use a local file make sure that the path is correct.
File: node_modules/got/index.js:91:20
not finished Building HTML renderer - 17.304s
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mapbox/mapbox-sdk",
-  "version": "0.15.2",
+  "version": "0.15.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


allanlewis:
I think is now redundant following the merge of #440.

PR #416, opened 2021-04-14
implement Fonts service

andrewharvey:
https://docs.mapbox.com/api/maps/fonts/

`getFontGlyphRange` still exists in Styles, which I suggest we do for backwards compatibility. However this means this method is now duplicate in the code base. So we can either:

- leave this duplicated, existing in both Fonts and Styles services
- omit it from the Fonts service
- mark as deprecated in Styles service, leaving duplicated code, but giving a path for future removal.

### DIFF
--- a/services/fonts.js
+++ b/services/fonts.js
@@ -1,0 +1,225 @@
+'use strict';
+
+var xtend = require('xtend');
+var v = require('./service-helpers/validator');
+var pick = require('./service-helpers/pick');
+var createServiceFactory = require('./service-helpers/create-service-factory');
+
+/**
+ * Fonts API service.
+ *
+ * Learn more about this service and its responses in
+ * [the HTTP service documentation](https://docs.mapbox.com/api/maps/fonts/).
+ */
+var Fonts = {};
+
+/**
+ * Get a font glyph range.
+ *
+ * See [the corresponding HTTP service documentation](https://docs.mapbox.com/api/maps/fonts/#retrieve-font-glyph-ranges).
+ *
+ * @param {Object} config
+ * @param {string|Array<string>} config.fonts - An array of font names.
+ * @param {number} config.start - Character code of the starting glyph.
+ * @param {number} config.end - Character code of the last glyph,
+ *   typically equivalent to`config.start + 255`.
+ * @param {string} [config.ownerId]
+ * @return {MapiRequest}
+ *
+ * @example
+ * fontsClient.getFontGlyphRange({
+ *   fonts: 'Arial Unicode',
+ *   start: 0,
+ *   end: 255
+ * })
+ *   .send()
+ *   .then(response => {
+ *     const glyph = response.body;
+ *   });
+ */
+Fonts.getFontGlyphRange = function(config) {
+  v.assertShape({
+    fonts: v.required(v.oneOfType(v.string, v.arrayOf(v.string))),
+    start: v.required(v.number),
+    end: v.required(v.number),
+    ownerId: v.string
+  })(config);
+
+  var fileName = config.start + '-' + config.end + '.pbf';
+
+  return this.client.createRequest({
+    method: 'GET',
+    path: '/fonts/v1/:ownerId/:fontList/:fileName',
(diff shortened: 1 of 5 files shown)


andrewharvey:
- [ ] I found in some cases I needed to pass fresh=true to listFonts, though at the moment I'm not sure if that forces the cache bypass, or I was just lucky that a new URL resulted in a cache miss and retrieved the latest font list. Needs more testing to understand if this query parameter is actually interpreted.

andrewharvey:
By the way, I needed this for https://gist.github.com/andrewharvey/ff5e2c15b8a60249a593dc189192eb02 which reports

- fonts used within an account
- which fonts each Style is using
- which Styles are using each font (useful for knowing which fonts you can clear out of your account to keep under the 100 fonts limit)
- which fonts Styles are using that aren't in your account (or standard Mapbox fonts) - useful if you've accidentally deleted a font from your account that was in use

andrewharvey:
@katydecorah interested in reviewing?

katydecorah:
👋 @andrewharvey I shared this PR with the team that works on Fonts API and they'll have someone take a look (hopefully this week)!

services/fonts.js
```diff
@@ -0,0 +1,225 @@
...
+ *   .send()
+ *   .then(response => {
+ *     const glyph = response.body;
+ *   });
+ */
+Fonts.getFontGlyphRange = function(config) {
+  v.assertShape({
+    fonts: v.required(v.oneOfType(v.string, v.arrayOf(v.string))),
+    start: v.required(v.number),
+    end: v.required(v.number),
```
jseppi, line 44: Since we know that `end` is always `start + 255`, we could make this method a little more ergonomic by only requiring the `start` parameter. Then, we'd just add `255` to it to come up with the correct fileName:

```js
var fileName = config.start + '-' + (config.start + 255) + '.pbf';
```

services/fonts.js
```diff
@@ -0,0 +1,225 @@
...
+ * @example
+ * fontsClient.putFont({
+ *   file: 'path/to/file.ttf'
+ * })
+ *   .send()
+ *   .then(response => {
+ *     const newFont = response.body;
+ *   });
+ */
+Fonts.putFont = function(config) {
```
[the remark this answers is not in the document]
andrewharvey: In docs/development.md it says to use "create" for POST requests, do you think it should still be addFont or createFont?

services/fonts.js
andrewharvey: Must end always equal start+255 or could users still request a larger slice?

services/fonts.js
```diff
@@ -0,0 +1,225 @@
...
+ * })
+ *   .send()
+ *   .then(response => {
+ *     const metadata = response.body;
+ *   });
+ */
+Fonts.updateFontMetadata = function(config) {
+  v.assertShape({
+    font: v.required(v.string),
+    metadata: v.required(v.plainObject),
```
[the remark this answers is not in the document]
andrewharvey: I guess that way is more strict as we can validate it more easily. If there were every any other metadata attributes which could be set in the future it might be better to leave them all bundled into a metadata object, but I'll make this change as suggested.

services/fonts.js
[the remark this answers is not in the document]
jseppi: 👍 `createFont` is a good suggestion for consistency

services/fonts.js
jseppi: Yup, `end` must always be `start + 255`

services/fonts.js
```diff
@@ -0,0 +1,223 @@
...
+ * })
+ *   .send()
+ *   .then(response => {
+ *     const metadata = response.body;
+ *   });
+ */
+Fonts.updateFontMetadata = function(config) {
+  v.assertShape({
+    font: v.required(v.string),
+    visibility: v.required.oneOf(['public', 'private']),
```
jseppi, line 211: Ah, sorry I got the syntax wrong in my review suggestion, @andrewharvey, which is why tests are failing. Should be:

```js
v.required(v.oneOf(['public', 'private']))
```

PR #422, opened 2021-06-22
Update got to 10.7.0

AmitPhulera:
There is a CVE in one of the dependencies that `got` uses. Can share more details in private.
Updating got to  10.7.0 fixes it. 
Assuming code has good test coverage, this would be a low-risk change. 
One test failed which I believe is a minor update in the `got` library. I have fixed it. 
Updating to the latest `got` version had more tests failing so I did not go that path.

It would be helpful if this can be taken up quickly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,9 +56,28 @@
       }
     },
     "@sindresorhus/is": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.7.0.tgz",
-      "integrity": "sha512-ONhaKPIufzzrlNbqtWFFd+jlnemX6lJAgq9ZeiZtS7I1PIf/la7CW4m83rTXRnVnsMbW2k56pGYu7AUFJD9Pow=="
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-2.1.1.tgz",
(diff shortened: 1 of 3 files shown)


AmitPhulera:
Hi @andrewharvey,
Thanks for the quick action, really appreciate it. 

Would it be possible for you to make releases for versions v0.11.1, v0.12.2, v0.13.1? 
I just want to be sure the changes make to https://github.com/mapbox/mapbox-gl-geocoder/

Thanks again for looking into it.

andrewharvey:
I don't think Mapbox would do patch releases like that, likely these changes would just flow through in the next release when that happens. If it's a security issue that meets Mapbox's eligibility requirements you can always report via https://www.mapbox.com/platform/disclosure/.

AmitPhulera:
> I don't think Mapbox would do patch releases like that, likely these changes would just flow through in the next release when that happens.

I was really hoping that these updates go in.

> If it's a security issue that meets Mapbox's eligibility requirements you can always report via https://www.mapbox.com/platform/disclosure/.

I am not sure about that and neither do I have the expertise to investigate further in this area.

I guess the only option I am left with is waiting until a new version of [Mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder/) comes out.

Thanks a lot for your response @andrewharvey 😄

andrewharvey:
You'd need to first do a release here, then do submit a PR to the geocoder to update the SDK-JS version, then do another release on the geocoder.

AmitPhulera:
I think making release here would fix geocoder as it uses "@mapbox/mapbox-sdk": "^0.11.0". 

https://github.com/mapbox/mapbox-gl-geocoder/blob/master/package.json#L62

So I guess having a v0.11.1 here would fix the issue.

PR #444, opened 2022-07-01
Added support for contours_meters Isochrone API param 

GrigoryDobrov:
mapbox/playground#193

There is a parameter `contours_meters` in official Isochrone API docs https://docs.mapbox.com/api/navigation/isochrone/ but it is not present in  mapbox-sdk-js. This PR adds support for `contour_meters` parameter.


services/isochrone.js
```diff
@@ -39,28 +41,47 @@ Isochrone.getContours = function(config) {
...
     throw new Error('minutes must be less than 60');
   }
 
+  if (
+    config.meters !== undefined &&
+    !config.meters.every(function(meter) {
+      return meter <= 100000;
+    })
+  ) {
+    throw new Error('meters must be less than 100000');
```
mikhail-kashirskii, line 82: can we have constant variable for this?

GrigoryDobrov: fixed

mpothier (approved):
LGTM

PR #440, opened 2022-06-22
Update `got` to 11.8.5 (CVE-2022-33987)

amureki:
Greetings, dear fellow maintainers,

Currently, there is a fixed CVE in the `got` package:
https://security.snyk.io/vuln/SNYK-JS-GOT-2932019

It is moderate according to GitHub:
https://github.com/advisories/GHSA-pfrx-2q88-qq97

And it is critical according to `npm audit`.

I am a user of Mapbox and using JS SDK (this one) in one of my projects and would like to get this fixed.
Therefore, I opened up this PR:

- Fix-up package-lock.json with new metadata
- Use node 14 in Travis CI (current LTS)
- Update `got` dependency to 11.8.5
- Update CHANGELOG.md

This seems bloated, but I tried to separate all steps that were done in separate commits.
I also checked release notes for https://github.com/sindresorhus/got/releases/tag/v11.0.0, it does not seem that things used here were affected, but it would be good for maintainers to confirm that.

And I am happy to reconsider things and try to simplify it if you see a better way.

Best,
Rust

PR #449, opened 2022-09-13
fix tests after got version bump

chriswhong:
Fixes two tests that broke after #440 was merged:

- `DELETE` request requires a body
- `progressUpload` returns a `0` instead of `undefined`


mpothier (approved):

mpothier:
Thanks @chriswhong , looks good to me.

PR #460, opened 2022-11-17
Add support for allowedApplications to Tokens.createToken and Tokens.updateToken

olemstrom:
Adds support for the new `allowedApplications` restriction to `Tokens.createToken` and `Tokens.updateToken`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- **Add:** Config for `Tokens#createToken` and `Tokens#updateToken` can now include the `allowedApplications` property.
+
+
 ## 0.13.7
 
 - **Add:** add additional EV values to the directions API request
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 0.14.0
+
 - **Add:** Config for `Tokens#createToken` and `Tokens#updateToken` can now include the `allowedApplications` property.
 
 
--- a/services/__tests__/tokens.test.js
+++ b/services/__tests__/tokens.test.js
@@ -73,6 +73,21 @@ describe('createToken', () => {
     });
   });
 
+  test('with allowedApplications', () => {
+    tokens.createToken({
+      allowedApplications: [{ platform: 'iOS', bundleId: 'com.example.foo' }]
+    });
+    expect(tu.requestConfig(tokens)).toEqual({
+      path: '/tokens/v2/:ownerId',
+      method: 'POST',
+      params: {},
+      body: {
+        allowedApplications: [{ platform: 'iOS', bundleId: 'com.example.foo' }],
+        scopes: []
+      }
+    });
+  });
+
   test('with scopes', () => {
     tokens.createToken({ scopes: ['styles:read', 'styles:write'] });
     expect(tu.requestConfig(tokens)).toEqual({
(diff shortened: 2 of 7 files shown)


services/tokens.js
```diff
@@ -61,7 +62,13 @@ Tokens.createToken = function(config) {
     note: v.string,
     scopes: v.arrayOf(v.string),
     resources: v.arrayOf(v.string),
-    allowedUrls: v.arrayOf(v.string)
+    allowedUrls: v.arrayOf(v.string),
+    allowedApplications: v.arrayOf(
+      v.shape({
+        bundleId: v.string,
+        platform: v.string
+      })
```
omidmaldar, line 70:
```suggestion
      v.strictShape({
        bundleId: v.string,
        platform: v.oneOf("Android", "iOS")
      })
```

I wonder do we want to prevent them from dispatching their request if platform is not `Android` or `iOS`?

services/tokens.js
```diff
@@ -150,7 +162,13 @@ Tokens.updateToken = function(config) {
...
+    )
```
omidmaldar, line 171:
```suggestion
    allowedApplications: v.arrayOf(
      v.strictShape({
        bundleId: v.string,
        platform: v.oneOf("Android", "iOS")
      })
    )
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- **Add:** Config for `Tokens#createToken` and `Tokens#updateToken` can now include the `allowedApplications` property.
```
omidmaldar, line 5:
```suggestion
## 0.13.8

- **Add:** Config for `Tokens#createToken` and `Tokens#updateToken` can now include the `allowedApplications` property.
```

This version number also should be also added to package.json and package-lock.json

services/__tests__/tokens.test.js
```diff
@@ -73,6 +73,21 @@ describe('createToken', () => {
...
+    tokens.createToken({
+      allowedApplications: [{ platform: 'iOS', bundleId: 'com.example.foo' }]
+    });
+    expect(tu.requestConfig(tokens)).toEqual({
+      path: '/tokens/v2/:ownerId',
+      method: 'POST',
+      params: {},
+      body: {
+        allowedApplications: [{ platform: 'iOS', bundleId: 'com.example.foo' }],
+        scopes: []
```
omidmaldar, line 86: I think it's fine if we drop `scopes` from this request body (?)

services/tokens.js
olemstrom: This is intentional as I don't want to encode every piece of validation logic into the SDK itself to make it easier to change in the backend.

CHANGELOG.md
olemstrom: Good point, I was thinking that package release is separate from the feature PR itself. I'll combine these 👍

omidmaldar: would be good to double-check this. Not sure, we want to release a public version just now, specially when this feature is not publicly available

services/__tests__/tokens.test.js
olemstrom: Empty scopes array is included in the request body even if its omitted from the SDK call itself. Thus it needs to be there for the assertion to pass.

CHANGELOG.md
olemstrom: This public version needs to exists for the sake of publishing the npm package

services/tokens.js
omidmaldar: Yep, potentially we will get more traffic because of Bad Requests but probably fine to gain more flexibility. Not sure what's the strategy here but let's move on. It's good enough for now.

omidmaldar (approved):

PR #461, opened 2023-01-06
Add gecoding v6 service

VladislavMedved:
## Description

Adds geocoding v6 service that supports forward, reverse and structured input modes

### DIFF
--- a/services/geocoding-v6.js
+++ b/services/geocoding-v6.js
@@ -111,31 +111,32 @@ var featureTypes = [
  *   });
  */
 GeocodingV6.forwardGeocode = function(config) {
-  v.assertShape({
-    query: v.required(v.string),
-    mode: v.oneOf('standard', 'structured'),
-    countries: config.mode === 'standard' ? v.arrayOf(v.string) : v.string,
-    proximity: v.oneOf(v.coordinates, 'ip'),
-    types: v.arrayOf(v.oneOf(featureTypes)),
-    bbox: v.arrayOf(v.number),
-    language: v.string,
-    limit: v.number,
-    worldview: v.string,
-    autocomplete: v.boolean,
-
-    // structured input fields
-    address_line1: v.string,
-    address_number: v.string,
-    street: v.string,
-    block: v.string,
-    place: v.string,
-    region: v.string,
-    neighborhood: v.string,
-    postcode: v.string,
-    locality: v.string
-  })(config);
-
   config.mode = config.mode || 'standard';
+
+  v.assertShape(
+    xtend(config.mode === 'standard' ? { query: v.required(v.string) } : {}, {
+      mode: v.oneOf('standard', 'structured'),
+      countries: config.mode === 'standard' ? v.arrayOf(v.string) : v.string,
+      proximity: v.oneOf(v.coordinates, 'ip'),
+      types: v.arrayOf(v.oneOf(featureTypes)),
+      bbox: v.arrayOf(v.number),
+      language: v.string,
+      limit: v.number,
+      worldview: v.string,
+      autocomplete: v.boolean,
+
+      // structured input fields
+      address_line1: v.string,
+      address_number: v.string,
+      street: v.string,
+      block: v.string,
+      place: v.string,
+      region: v.string,
+      neighborhood: v.string,
+      postcode: v.string,
+      locality: v.string
+    })
+  )(config);
 
   var query = stringifyBooleans(
     xtend(
--- a/services/geocoding-v6.js
+++ b/services/geocoding-v6.js
@@ -1,0 +1,231 @@
+'use strict';
+
+var xtend = require('xtend');
+var v = require('./service-helpers/validator');
+var pick = require('./service-helpers/pick');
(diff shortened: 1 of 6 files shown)


services/geocoding-v6.js
```diff
@@ -0,0 +1,231 @@
...
+    limit: v.number,
+    worldview: v.string,
+    autocomplete: v.boolean,
+
+    // structured input fields
+    address_number: v.string,
+    street: v.string,
+    place: v.string,
+    region: v.string,
+    country: v.string,
```
mpothier, line 127: `country` and `countries` seem somewhat redundant. Do we need them both for different modes, or can they be consolidated?

services/geocoding-v6.js
```diff
@@ -0,0 +1,231 @@
...
+    countries: v.arrayOf(v.string),
+    proximity: v.oneOf(v.coordinates, 'ip'),
+    types: v.arrayOf(v.oneOf(featureTypes)),
+    bbox: v.arrayOf(v.number),
+    language: v.string,
+    limit: v.number,
+    worldview: v.string,
+    autocomplete: v.boolean,
+
+    // structured input fields
```
mpothier, line 122: It looks like this is missing `address_line1`

services/geocoding-v6.js
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * Geocoding API service.
+ *
+ * Learn more about this service and its responses in
+ * [the HTTP service documentation](https://docs.mapbox.com/api/search/geocoding-v6/).
+ */
+var GeocodingV6 = {};
+
+var featureTypes = [
```
mpothier, line 17: Confirm if we have `street` as a feature type or not

services/geocoding-v6.js
```diff
@@ -0,0 +1,231 @@
...
+    address_number: v.string,
+    street: v.string,
+    place: v.string,
+    region: v.string,
+    country: v.string,
+    postcode: v.string,
+    locality: v.string
+  })(config);
+
+  config.mode = config.mode || 'normal';
```
mpothier, line 132: I'd propose using a different word for 'normal'. Some ideas are `'simple'`, '`standard'`

services/geocoding-v6.js
```diff
@@ -0,0 +1,231 @@
...
+ *   .then(response => {
+ *     const match = response.body;
+ *   });
+ */
+GeocodingV6.forwardGeocode = function(config) {
+  v.assertShape({
+    query: v.required(v.string),
+    mode: v.oneOf('normal', 'structured'),
+    countries: v.arrayOf(v.string),
+    proximity: v.oneOf(v.coordinates, 'ip'),
```
mpothier, line 114: Can you check what object-type is referred to by `v.coordinates`? i.e. does it expect a number array, or does it accept an object with `latitude` and `longitude` properties?

mpothier (changes requested):

services/geocoding-v6.js
VladislavMedved: @krydima confirmed we have `street` feature type

services/geocoding-v6.js
VladislavMedved: it's an array of numbers `v.tuple(v.number, v.number)`

services/geocoding-v6.js
VladislavMedved: I don't think we need it both, we can try to consolidate it to one `countries` parameter

VladislavMedved: I've also added `block` and `neighborhood` from v6 api documentation

services/geocoding-v6.js
```diff
@@ -0,0 +1,239 @@
...
+ *   bbox: [2.14, 48.72, 2.55, 48.96]
+ * })
+ *   .send()
+ *   .then(response => {
+ *     const match = response.body;
+ *   });
+ */
+GeocodingV6.forwardGeocode = function(config) {
+  v.assertShape({
+    query: v.required(v.string),
```
mpothier, line 115: @VladislavMedved will `query` be required even if intending to use structured mode? Will it throw a validation error if `query` is not included?

mpothier (approved):
@VladislavMedved this looks good, with my only question about whether an empty string for `query` is required to pass validation when using structured mode.

services/geocoding-v6.js
VladislavMedved: Good catch! I've made it required only for `standard` mode

services/geocoding-v6.js
```diff
@@ -0,0 +1,243 @@
...
+ *     const match = response.body;
+ *   });
+ */
+GeocodingV6.reverseGeocode = function(config) {
+  v.assertShape({
+    longitude: v.required(v.number),
+    latitude: v.required(v.number),
+    countries: v.arrayOf(v.string),
+    types: v.arrayOf(v.oneOf(featureTypes)),
+    bbox: v.arrayOf(v.number),
```
mpothier, line 215: Please double check if `bbox` applies for reverse queries.

PR #465, opened 2023-04-27
add driving-traffic profile to isochone api service

chriswhong:
Adds `driving-traffic` profile to the isochrone service


walkerke:
@chriswhong @Countdown369 have you all at Mapbox considered re-enabling the `depart_at` parameter for the Isochrone API service?  That was a really nice feature that appears to have been discontinued.

PR #468, opened 2023-05-23
add directions api parameters in support of route notifications

chriswhong:
Adds `max_width`, `max_height`, `max_weight` and `notifications` parameters to the directions service


Countdown369 (approved):
LGTM

Issue #503, opened 2025-10-31
Upgrade form-data library version to 4.0.4 due to security advisory


doug-numetric:
Type|url
-- | --
Original advisory | https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2025/07/GHSA-fjxv-7rqg-78g4/GHSA-fjxv-7rqg-78g4.json
Web | https://github.com/form-data/form-data/security/advisories/GHSA-fjxv-7rqg-78g4
Advisory | https://nvd.nist.gov/vuln/detail/CVE-2025-7783
Web | https://github.com/form-data/form-data/commit/3d1723080e6577a66f17f163ecd345a21d8d0fd0
Web | https://github.com/benweissmann/CVE-2025-7783-poc

PR #506, opened 2026-05-01
Upgrade jest-environment-jsdom

mmarkov:
Starting Jest 28 global becomes accessible without need of `jest-environment-jsdom-global`. So I am removing `jest-environment-jsdom-golobal` and as result we are getting rid dependency on `form-data 4.0.0`

This change should address issue - https://github.com/mapbox/mapbox-sdk-js/issues/503


mpothier (approved):
LGTM

